PR #1, opened 2021-10-15
Unit Testing

jvyden:
This branch adds unit tests, which can be very useful for a project like this. I figured it would be good to get this done early, so here we are.

I'm making this merge request because I've never done unit testing in C# before and just wanted to make sure this was a decent way of doing it.

I've already found a bug (see 4bce25e) with this, so it'll definitely be useful in the long run


jvyden:
Had a friend (@jiniux) take a look, they said it looked fine, merging.

PR #35, opened 2021-11-02
Profile pins

LumaLivy:
Adds support for setting profile pins.

### Some notes:
- Tested in LBP2, RPCS3 environment
- Added a new class under Profile in Types for "Pins", because I needed a class in order to deserialize the JSON request body being posted. Hopefully this is the appropriate place to put it.
- If there is a better way to grab the profile_pins key from the `update_my_pins` request without the Pins class, or the class needs more (like the `Serialize` method or xml attributes seen in other "Types"), please let me know. I'm familiarizing myself with C# so I picked a relatively simple task to undertake.
- I'm not sure if the progress and awards fields are used for online play because the `profile_pins` are what's being set in the actual profile. No database changes are needed for my additions because the joined `profile_pins` string is being set in `User.Pins`

### DIFF
--- a/ProjectLighthouse/Types/Profiles/Pins.cs
+++ b/ProjectLighthouse/Types/Profiles/Pins.cs
@@ -1,12 +1,14 @@
-using LBPUnion.ProjectLighthouse.Serialization;
-
+using System.Text.Json.Serialization;
 namespace LBPUnion.ProjectLighthouse.Types.Profiles
 {
     public class Pins
     {
-        public long[] progress { get; set; }
-        public long[] awards { get; set; }
-        public long[] profile_pins { get; set; }
+        [JsonPropertyName("progress")]
+        public long[] Progress { get; set; }
+        [JsonPropertyName("awards")]
+        public long[] Awards { get; set; }
+        [JsonPropertyName("profile_pins")]
+        public long[] ProfilePins { get; set; }
     }
 }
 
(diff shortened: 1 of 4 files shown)


ProjectLighthouse/Types/Profiles/Pins.cs
```diff
@@ -0,0 +1,12 @@
+using LBPUnion.ProjectLighthouse.Serialization;
+
+namespace LBPUnion.ProjectLighthouse.Types.Profiles
+{
+    public class Pins
+    {
+        public long[] progress { get; set; }
+        public long[] awards { get; set; }
+        public long[] profile_pins { get; set; }
```
jvyden, line 9: You can keep the naming style of the JSON data while also keeping the naming style of C# using a `JsonPropertyName` attribute.
Heres a (modified) example [from the Microsoft wiki](https://docs.microsoft.com/en-us/dotnet/standard/serialization/system-text-json-customize-properties?pivots=dotnet-5-0#customize-individual-property-names):
```csharp
[JsonPropertyName("wind_speed")]
public int WindSpeed { get; set; }
```

jvyden (changes requested):

jvyden:
Pretty good! Just one thing, as showed above.

jvyden:
@github-actions whats your deal run the damn tests

LumaLivy:
That should do it. Thanks for the pointer!

jvyden:
Looks good to me, thanks for your contribution!

TorutheRedFox:
the progress and awards are used to award specific pins, and seemingly to carry pins over from LBP2 to LBP3?

Issue #22, opened 2021-10-25
MM Picks / Lucky Dip Pages


jvyden:

jvyden:
Implemented in #36.

PR #36, opened 2021-11-03
Add support for lucky dipping in LBP2

LumaLivy:
In short, it functions a lot like grabbing the MM picks. The endpoint is specifically tied to LBP2, so the other games will need to be tested for their endpoints. Referencing #22 so we can work on achieving feature parity.

It would be good if someone with access to both LBP1 and LBP3 can test what their endpoints look like. Also, someone with access to LBP1 level publishing should verify if this can pull up LBP1 published levels from inside LBP2. I don't see why it shouldn't, but then again I can't test it on RPCS3.

### Final notes 
The game includes a seed in the url (seems like a number 1-99) for randomizing what levels are grabbed (along with the selected filters etc). I had trouble incorporating `System.Random` with a seed into the `OrderBy` because it couldn't be translated to SQL, but generating a random guid worked for randomness, so I implemented it that way. Depending on how big the database gets this may become a performance hit ordering everything randomly and then grabbing the top 30 but I don't really think we're looking at hosting millions of levels in one instance!


jvyden (approved):
Looks good to me, thanks for your contribution.

PR #37, opened 2021-11-04
Add support for grabbing Top Scores

LumaLivy:
Lots going on here, so I'll break down the decisions I made/things I've done:

### LbpSerializer.cs
- Added another definition of TaggedStringElement which allows providing a dictionary of key/value pairs, for cases were an XML tag has multiple attributes. You can see how I use it when sending the `<scores />` tag to the game.

### ScoreController.cs

- Fixed a bug I noticed when submitting scores that would submit a new score by the same person each time a level is completed, rather than update their best highscore. I'm aware that there's probably something database related that could be done to enforce a unique constraint on the `PlayerIdCollection` but for now I just did the checking all in code.
- Added endpoint for `topscores`

Most of the `TopScores` code should be straight forward with the comments. I did my best to try to compute everything in the server but I had issues with LINQ not translating the `SELECT` part of the `rankedScores` query which pairs up the ordered results with a rank number, so I gave in with `ToList()`. It's also worth mentioning it's important I do the ranking there because the game will pin the player's own score + rank above the top scores list -- so without knowing where they rank overall (before cutting down the results for pagination) the user won't know their true rank.

I don't like having to use `var` but unfortunately the `SELECT` returns an Anonymous Type which made me pull my hair out for hours. I tried an approach where I had a new class for RankedScore but it still wouldn't translate with LINQ or anything so I scrapped it all and kept the anonymous return. Any suggestions are welcome.

### Score.cs

Finally, I added some more properties to support top scores, as well as a way to serialize a `playRecord`. From what I can tell, main player takes the credit for a group score in multiplayer, so I just set it to whatever the first value of the PlayerIdCollection is. Rank ends up getting attached while loading the top scores list.

### Conclusion

Phew! What a report. Hopefully this lays some more of the groundwork for scoring. I would like to take a look at scores for the story levels as well, because they use a different endpoint. It's very possible most of this code will be recycled.

### DIFF
--- a/ProjectLighthouse.Tests/Tests/AuthenticationTests.cs
+++ b/ProjectLighthouse.Tests/Tests/AuthenticationTests.cs
@@ -12,7 +12,7 @@ namespace LBPUnion.ProjectLighthouse.Tests
         [Fact]
         public async Task ShouldReturnErrorOnNoPostData()
         {
-            HttpResponseMessage response = await this.Client.PostAsync($"/LITTLEBIGPLANETPS3_XML/login", null!);
+            HttpResponseMessage response = await this.Client.PostAsync("/LITTLEBIGPLANETPS3_XML/login", null!);
             Assert.False(response.IsSuccessStatusCode);
             #if NET6_0_OR_GREATER
             Assert.True(response.StatusCode == HttpStatusCode.BadRequest);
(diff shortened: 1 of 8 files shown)


jvyden (approved):

jvyden:
Yeah, we can definitely reuse code for story levels. Story levels are typically referred to using "developer" slots instead of the typical "user" slots. It shouldn't be too hard to implement that.

PR #38, opened 2021-11-08
Add Playcount support, includes various bugfixes/other updates

LumaLivy:
Main contribution for this massive PR is playcount support (sorry, it was a smaller PR originally but I accidentally merged the playcount changes into my main branch and it automatically tacked it onto this PR).

### TLDR
I updated this PR comment to reflect the stuff I added for playcount support, consider this more of a double PR lol.

### Playcount

The main takeaway for playcount support was adding the appropriate columns to Slots, and creating a table for VisitedLevels which is just a way to keep track of if a user has entered a level for the first time. If so, their play is counted as unique and increments the play counter shown on the level thumbnail ingame. If not, their play is still counted as a regular play in the level statistics. Submitting a score triggers incrementing the completion count. Mostly self explanatory stuff for the rest.

### -- old PR below --
Lots of small things going on past this point and didn't want to deviate too much from main so I thought I'd better PR. Very quick summary of each file:

### ScoreController.cs
- make `GetScores()` that is meant to be a reusable way of grabbing scores from the database
  - by default, if no page parameters are provided, it assumes you want to find the user's score in the DB (page unknown), so it grabs the 2 scores before and the 2 after, and the user's score. Hoping this will be used when I find a way to show scores on the 2nd page of the scoreboard.
  - if every parameter is provided, it pulls the global top scores list with the page parameters
  - later, this will also be used to grab friend scores

### UserController.cs
- Same idea as ScoreController.cs, but make `GetSerializedUser()` to have a reusable way to grab serialized users. Going to come in handy later with streams/need for grabbing users without reusing code.
- Let the users endpoint grab a list of users properly, the game does this with query strings for multiple users that look like this: ```users?u=LumaLivy&u=EvilLumaLivy&u=user1```

### MatchHelper.cs
- Fixed a bug where MM was sending a hex for the location IP which is illegal JSON, causing the JSON deserializer to choke. I simply do a regex match to convert any hex numbers to int32

### LighthouseFileLogger
- Added a try/catch because windows was complaining about resources being in use when trying to append to the log file. Who knows.

### CreateRoom.cs
- I made this file locally at the same time as the other that was added earlier, but I added more fields that I saw pop up on my end. I figured I'd just leave them in and merge them with what's on main. Not sure how much of them are needed for now.


ProjectLighthouse/Controllers/ScoreController.cs
```diff
@@ -49,20 +59,31 @@ public async Task<IActionResult> SubmitScore(int id)
...
             await this.database.SaveChangesAsync();
 
-            return this.Ok();
+            string myRanking = await GetScores(score.SlotId, score.Type, user);
+
+            return this.Ok(myRanking);
         }
 
+        [HttpGet("friendscores/user/{slotId:int}/{type:int}")]
+        public async Task<IActionResult> FriendScores(int slotId, int type)
```
jvyden, line 71: Since theres no async calls:

```suggestion
        public IActionResult FriendScores(int slotId, int type)
```

ProjectLighthouse/Controllers/ScoreController.cs
```diff
@@ -49,20 +59,31 @@ public async Task<IActionResult> SubmitScore(int id)
...
+        //=> await TopScores(slotId, type);
+        => this.Ok("<scores />");
```
jvyden, line 73:
```suggestion
        => this.Ok("<scores/>");
```
or:
```suggestion
        => this.Ok(LbpSerializer.BlankElement("scores"));
```

ProjectLighthouse/Controllers/UserController.cs
```diff
@@ -24,18 +25,32 @@ public UserController(Database database)
...
-
-            return this.Ok(user.Serialize());
+            return this.Ok(user);
         }
 
         [HttpGet("users")]
-        public async Task<IActionResult> GetUserAlt([FromQuery] string u) => await this.GetUser(u);
+        public async Task<IActionResult> GetUserAlt([FromQuery] string[] u)
+        {
+            List<string> tasks = new();
```
jvyden, line 44: `tasks`?

ProjectLighthouse/Controllers/UserController.cs
```diff
@@ -24,18 +25,32 @@ public UserController(Database database)
...
+        public async Task<string> GetSerializedUser(string username)
         {
-            User user = await this.database.Users.Include(u => u.Location).FirstOrDefaultAsync(u => u.Username == username);
+            User? user = await this.database.Users.Include(u => u.Location).FirstOrDefaultAsync(u => u.Username == username);
+            return user?.Serialize();
+        }
 
+        [HttpGet("user/{username}")]
+        public async Task<IActionResult> GetUser(string username) {
+            string? user = await this.GetSerializedUser(username);
```
jvyden, line 36: This file should have a `#nullable enable` at the beginning of it if you use nullable types.

ProjectLighthouse/Migrations/20211108100949_GameSpecificPlayCountsRedo.cs
```diff
@@ -0,0 +1,17 @@
...
+        {
+
+        }
+
+        protected override void Down(MigrationBuilder migrationBuilder)
+        {
+
+        }
+    }
+}
```
jvyden, line 17: This whole file can be removed since it doesn't do anything.

ProjectLighthouse/Types/Levels/Slot.cs
```diff
@@ -52,6 +52,12 @@ public class Slot
         [ForeignKey(nameof(CreatorId))]
         public User Creator { get; set; }
 
+        [NotMapped]
+        public string CreatorName
+        {
+            get => this.Creator?.Username;
+        }
+
```
jvyden, line 60: I don't really see the point of this...

jvyden (changes requested):

ProjectLighthouse/Controllers/UserController.cs
LumaLivy: Old variable name, my previous strategy was running async calls in parallel but I learned you can't access a single DbContext from multiple threads at the same time. The choices were to have a context for each async call or make it synchronous, so I went with synchronous and forgot to update the name to indicate I wasn't getting a list of tasks anymore. Updating this for clarity in my next commit.

ProjectLighthouse/Types/Levels/Slot.cs
jvyden: You need to include the creator in the query like so:

```csharp
this.database.Slots.Include(s => s.Creator)
```

ProjectLighthouse/Migrations/20211108100949_GameSpecificPlayCountsRedo.cs
LumaLivy: Fixed in https://github.com/LBPUnion/project-lighthouse/pull/38/commits/f93c9c5e329d9f2b8113f1d28015fd979ba0a209

ProjectLighthouse/Controllers/UserController.cs
LumaLivy: Fixed in https://github.com/LBPUnion/project-lighthouse/pull/38/commits/45b8464d1cb676080d24e3cc5a4036ff65d804cb

ProjectLighthouse/Types/Levels/Slot.cs
jvyden: It doesn't include it automatically because it (rightfully) assumes you don't need it by default.

ProjectLighthouse/Controllers/UserController.cs
LumaLivy: Fixed in https://github.com/LBPUnion/project-lighthouse/pull/38/commits/45b8464d1cb676080d24e3cc5a4036ff65d804cb

ProjectLighthouse/Types/Levels/Slot.cs
LumaLivy: Made things look more like the original code but with the nullable on Creator, see https://github.com/LBPUnion/project-lighthouse/pull/38/commits/ac8e1dc303f2b2d408969b8eceafe838a013b020

LumaLivy: Found the culprit. See https://github.com/LBPUnion/project-lighthouse/pull/38/commits/7e294b8c2110dca1fb88dd4f37a17bb33685fb50 for the missing include for queued levels.
Fixed!

jvyden (approved):

PR #39, opened 2021-11-09
Add support for level ratings

LumaLivy:
Self explanatory, lets a user Yay and Boo levels, as well as send a 5 star rating of a level in LBP1. 

A user that has played a level in both LBP1 and LBP2+ can simultaneously yay/boo a level and have a numeric LBP1 rating associated to them.

### DIFF
--- a/ProjectLighthouse/Controllers/SlotsController.cs
+++ b/ProjectLighthouse/Controllers/SlotsController.cs
@@ -45,18 +45,27 @@ namespace LBPUnion.ProjectLighthouse.Controllers
         [HttpGet("s/user/{id:int}")]
         public async Task<IActionResult> SUser(int id)
         {
+            User? user = await this.database.UserFromRequest(this.Request);
+            if (user == null) return this.StatusCode(403, "");
+
             Token? token = await this.database.TokenFromRequest(this.Request);
             if (token == null) return this.BadRequest();
 
             GameVersion gameVersion = token.GameVersion;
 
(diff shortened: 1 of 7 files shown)


ProjectLighthouse/Controllers/SlotsController.cs
```diff
@@ -45,18 +45,27 @@ public async Task<IActionResult> SlotsBy([FromQuery] string u)
...
 
             if (slot == null) return this.NotFound();
 
+            RatedLevel? ratedLevel = await this.database.RatedLevels.FirstOrDefaultAsync(r => r.SlotId == id && r.UserId == user.UserId);
+            if (ratedLevel != null) {
+                slot.YourRating = ratedLevel.RatingLBP1;
+                slot.YourDPadRating = ratedLevel.Rating;
+            }
+            
             return this.Ok(slot.Serialize());
```
jvyden, line 69: not entirely sure setting it on the slot directly is a good idea

maybe pass it in to `Slot.Serialize()` or make it a parameter similar to `Slot.Hearts`?

jvyden (changes requested):

ProjectLighthouse/Controllers/SlotsController.cs
LumaLivy: Fixed in https://github.com/LBPUnion/project-lighthouse/pull/39/commits/f157647f689ae687bbb52a6b5b4e922e0d0083ee

jvyden (approved):
prolly fine

PR #40, opened 2021-11-10
Update SlotsController.cs

hydronePVP45:
change cool levels to show newest levels instead of an unfunctional lucky dip


jvyden:
Code is fine, but why?

PR #50, opened 2021-11-14
Add GameVersion.LittleBigPlanetVita support

hydronePVP45:


jvyden:
It would probably be better to add Slot.PlaysLBPVita, since LBP2 and LBPVita are different games.

jvyden:
Done in #55.

PR #56, opened 2021-11-15
Added Pages on Newest Levels, Lucky Dip and Team Picks

hydronePVP45:


ProjectLighthouse/Controllers/SlotsController.cs
```diff
@@ -99,6 +99,8 @@ public async Task<IActionResult> NewestSlots([FromQuery] int pageStart, [FromQue
 
             GameVersion gameVersion = token.GameVersion;
 
+            int totalSlotCount = await this.database.Slots.CountAsync();
```
jvyden, line 102: This is only used once, so just put it in the dictionary.

ProjectLighthouse/Controllers/SlotsController.cs
```diff
@@ -118,6 +130,8 @@ public async Task<IActionResult> TeamPickedSlots([FromQuery] int pageStart, [Fro
 
             GameVersion gameVersion = token.GameVersion;
 
+            int mmPicksCount = await this.database.Slots.CountAsync(s => s.TeamPick);
```
jvyden, line 133: This is only used once, so just put it in the dictionary.

ProjectLighthouse/Controllers/SlotsController.cs
```diff
@@ -138,6 +161,8 @@ public async Task<IActionResult> LuckyDipSlots([FromQuery] int pageStart, [FromQ
 
             GameVersion gameVersion = token.GameVersion;
 
+            int totalSlotCount = await this.database.Slots.CountAsync();
```
jvyden, line 164: This is only used once, so just put it in the dictionary.

jvyden (changes requested):

PR #62, opened 2021-11-20
Add (base) support for reviews

LumaLivy:
### Nearly comprehensive review implementation, in LBP2 everything is functional, including:
- Posting/editing reviews
- Viewing/rating reviews both in-level and in-pod (both on the level page, and on the author's profile)
-  Reviews deleted by the level author will show that they have been deleted by the level author. Technically also supports moderator deletions as well, but there is no way to delete a review like that from in-game

### The Big Caveat:
The "leave a review!" button does not show up in a level, or in the pod. I have no idea why and I've spent like two days trying to solve this issue. Reviews can be manually inserted into the DB with no issues. Things I have tried to get the post button to appear ingame:
- passing `<reviewsEnabled>true</reviewsEnabled>` to slots (this is an expected value that should be on, just not user-configurable for now)
- guessing a few random values that would be similar in name to the others in `network_settings.nws` (I also searched in the EBOOT for the keys in this file and found them, but we don't seem to be missing anything)
- various other things like adding more missing values to users/slots in a hardcoded way, in case the game thinks something is wrong with your profile and prevents you from reviewing (idk). Nothing seemed to make much of a difference

### So...

Before I come up with some janky hack (for example, *editing* your existing review works fine... maybe I'll have the server return an empty review for you if there isn't one in the DB after a play) I thought it would still be worth pushing these changes to main in their current state. There are a few very minor bug fixes hidden in here as well that I noticed, such as profile comments not counting correctly, they've been fixed.


ProjectLighthouse/Controllers/EnterLevelController.cs
```diff
@@ -118,6 +119,7 @@ public async Task<IActionResult> EnterLevel(int id)
             {
                 v = await visited.FirstOrDefaultAsync();
             }
+            if (v == null) return StatusCode(403, "");
 
             if (v == null)
             {
```
jvyden, line 125: Not sure why this is being checked and returned twice.

ProjectLighthouse/Controllers/EnterLevelController.cs
```diff
@@ -62,6 +62,7 @@ public async Task<IActionResult> PlayLevel(int slotId)
...
                 v = await visited.FirstOrDefaultAsync();
             }
+            if (v == null) return this.StatusCode(403, "");
 
             if (v == null)
             {
```
jvyden, line 68: Not sure why this is being checked and returned twice.

ProjectLighthouse/Controllers/ReviewController.cs
```diff
@@ -74,5 +81,177 @@ public async Task<IActionResult> DPadRate(int slotId, [FromQuery] int rating)
...
+            GameVersion gameVersion = token.GameVersion;
+
+            Random rand = new();
+
+            IEnumerable<Review> reviews = this.database.Reviews.Where(r => r.SlotId == slotId && r.Slot.GameVersion <= gameVersion)
+                .Include(r => r.Reviewer)
+                .Include(r => r.Slot)
+                .AsEnumerable() // performance? Needed for next line (ThumbsUp is not in DB)
+                .OrderByDescending(r => r.ThumbsUp)
+                .ThenByDescending(_ => rand.Next())
```
jvyden, line 141: Hmm, it's a hack but we might be able to store and cache a list of reviews and their ThumbsUp values and send that to the DB that way. Otherwise we're left through ordering the entire table like this.

ProjectLighthouse/Controllers/ReviewController.cs
```diff
@@ -74,5 +81,177 @@ public async Task<IActionResult> DPadRate(int slotId, [FromQuery] int rating)
...
+            Token? token = await this.database.TokenFromRequest(this.Request);
+            if (token == null) return this.StatusCode(403, "");
+
+            GameVersion gameVersion = token.GameVersion;
+
+            IEnumerable<Review> reviews = this.database.Reviews.Where(r => r.Reviewer.Username == username && r.Slot.GameVersion <= gameVersion)
+                .Include(r => r.Reviewer)
+                .Include(r => r.Slot)
+                .AsEnumerable() // performance?
+                .OrderByDescending(r => r.Timestamp)
```
jvyden, line 179: Why cant this be ordered server-side?

ProjectLighthouse/Controllers/ReviewController.cs
```diff
@@ -74,5 +81,177 @@ public async Task<IActionResult> DPadRate(int slotId, [FromQuery] int rating)
...
+        [HttpPost("rateReview/user/{slotId:int}/{username}")]
+        public async Task<IActionResult> RateReview(int slotId, string username, [FromQuery] int rating = 0) {
+            User? user = await this.database.UserFromRequest(this.Request);
+            if (user == null) return this.StatusCode(403, "");
+
+            User? reviewer = await this.database.Users.FirstOrDefaultAsync(u => u.Username == username);
+            if (reviewer == null) return this.StatusCode(403, "");
+
+            Review? review = await this.database.Reviews.FirstOrDefaultAsync(r => r.SlotId == slotId && r.ReviewerId == reviewer.UserId);
+            if (review == null) return this.StatusCode(403, "");
```
jvyden, line 211: These should be returning 400/404

ProjectLighthouse/Controllers/ReviewController.cs
```diff
@@ -74,5 +81,177 @@ public async Task<IActionResult> DPadRate(int slotId, [FromQuery] int rating)
...
+        [HttpPost("deleteReview/user/{slotId:int}/{username}")]
+        public async Task<IActionResult> DeleteReview(int slotId, string username) {
+            User? reviewer = await this.database.Users.FirstOrDefaultAsync(u => u.Username == username);
+            if (reviewer == null) return this.StatusCode(403, "");
+
+            Review? review = await this.database.Reviews.FirstOrDefaultAsync(r => r.SlotId == slotId && r.ReviewerId == reviewer.UserId);
+            if (review == null) return this.StatusCode(403, "");
+
+            review.Deleted = true;
+            review.DeletedBy = "level_author"; // other value is "moderator"
```
jvyden, line 239: Can we have an enum for this for simplicity's sake?

ProjectLighthouse/Controllers/UserController.cs
```diff
@@ -39,7 +40,7 @@ public async Task<IActionResult> GetUser(string username)
             if (token == null) return this.StatusCode(403, "");
 
             string? user = await this.GetSerializedUser(username, token.GameVersion);
-            if (user == null) return this.NotFound();
+            if (user == "") return this.NotFound();
```
jvyden, line 43: `if(string.isNullOrEmpty(user))`

jvyden (changes requested):

ProjectLighthouse/Controllers/EnterLevelController.cs
LumaLivy: thought I pushed the commit to fix this, fixed in https://github.com/LBPUnion/project-lighthouse/pull/62/commits/a7443c16a0e240b50b4026f98ad60d5e744fc6a5

ProjectLighthouse/Controllers/EnterLevelController.cs
LumaLivy: ditto, https://github.com/LBPUnion/project-lighthouse/pull/62/commits/a7443c16a0e240b50b4026f98ad60d5e744fc6a5

ProjectLighthouse/Controllers/ReviewController.cs
LumaLivy: You're right, I think it came from something I copied and pasted. Fixed https://github.com/LBPUnion/project-lighthouse/pull/62/commits/abfb1c4cd86f0a222ad14f148d7a1e9931e83444

ProjectLighthouse/Controllers/ReviewController.cs
LumaLivy: I went with 400, makes sense to me that a bad request would be trying to rate something that doesn't exist. https://github.com/LBPUnion/project-lighthouse/pull/62/commits/ed1eac696a11a3bfb2b477b0bfec4c153fcafecf

ProjectLighthouse/Controllers/UserController.cs
LumaLivy: TIL, that's handy. Thanks. https://github.com/LBPUnion/project-lighthouse/pull/62/commits/d36a7d576560517cecfc6260661c681a5caf82be

ProjectLighthouse/Controllers/ReviewController.cs
LumaLivy: Well, I also learned today that C# can't have string enums so this is close enough, works practically the same way. https://github.com/LBPUnion/project-lighthouse/pull/62/commits/ece00c218ede04b4226bf45bcc27fc4449c4e557

I opted not to namespace it with reviews specifically because comments (both profile and level) can be deleted and I thought it would be a bit redundant to have separate types for each.

PR #102, opened 2021-12-11
Support Level Reviews (without mucking up the main repo)

LumaLivy:
Okay, so this is a bit messy but this is my second attempt at this pull request after the last one got merged by accident

### What's not new

Almost everything included in the additions, most of the additions were patched back in manually from the revert on main when the PR merged

### What is new
- Some funky new auto styling and a bunch of junk files added to the .gitignore that vscode created while I was merging changes (lots of moving curly braces to a new line etc., creating a bunch of logs which were missing from the .gitignore from some reason)
- DeletedBy is now a proper enum that serializes into XML
  - Warning: the code that does the serialization of the enum is kinda gross with `XmlSerializer` and `StringWriter`, I'm talking about [these lines](https://github.com/LumaLivy/project-lighthouse/blob/16a41d4fb50e1506604664e22fe6c83ccd03eddd/ProjectLighthouse/Types/Reviews/Review.cs#L72-L81). Haven't done this before so I used what I could find online. Most of it was done this way because the xml serializer would insert the `<?xml>` header thing in front of the root node and I didn't want to write another serialize method for the enum (lol)
- Review stats (thumbs up/down etc) are now pulled from the review itself ("cached" in database), so when someone rates the review, a review rating is created and the thumb count is updated on the review directly (the RatedReviews table exists so that we can track if a user has previously voted on a review before, they may rescind their rating or update it later). If I have the right approach, this can start being implemented for some of the other heavy DB aggregation calls (create a visit activity and increment the level playcount by 1 when you visit for both tracking and counting purposes, so when getting playcount of a slot while browsing, we can pull the number right from the row in the database without having to count the number of visits for that slot in the visit table, etc).


LumaLivy:
Oh, while I'm thinking of it, I will add some known issues that I observed:

- The add review button is still nowhere to be found, so my workaround is giving a user a blank review that is only visible to them  once they've played the level (and it is not their own). The template prompts them to edit it, and doing so will upload the review.
- Editing an existing review will show the user a blank review instead of their existing review (so clicking cancel fast will clear the review in the database). Needs a closer look as to why the game isn't passing the existing info to the review editor.
- Likewise, rating a review will hover over the boo option by default regardless of the user's existing yay/boo on a review.

ProjectLighthouse/Controllers/ReviewController.cs
```diff
@@ -69,10 +76,242 @@ public async Task<IActionResult> DPadRate(int slotId, [FromQuery] int rating)
...
+
+                yourReview = new();
+                yourReview.ReviewerId = user.UserId;
+                yourReview.Reviewer = user;
+                yourReview.Thumb = ratedLevel?.Rating == null ? 0 : ratedLevel.Rating;
+                yourReview.Slot = slot;
+                yourReview.SlotId = slotId;
+                yourReview.Deleted = false;
+                yourReview.DeletedBy = DeletedBy.None;
+                yourReview.Text = "You haven't reviewed this level yet. Edit this blank review to upload one!";
```
jvyden, line 168: beautifully scuffed

ProjectLighthouse/Controllers/ReviewController.cs
```diff
@@ -69,10 +76,242 @@ public async Task<IActionResult> DPadRate(int slotId, [FromQuery] int rating)
...
+                yourReview.LabelCollection = "";
+                yourReview.Timestamp = TimeHelper.UnixTimeMilliseconds();
```
jvyden, line 170: You can also do something like this:
```csharp
yourReview = new {
    ReviewerId = user.UserId,
    Reviewer = user,
    ...
};
```

jvyden (approved):

jvyden:
I'll resolve conflicts.

PR #105, opened 2021-12-14
New authentication system

### LINKED ISSUE #79: Remember IP address (auth)
The website should have an option to remember an IP address in the auth section so you do not have to press accept every time you log in.

jvyden:
This new method no longer lets the user in at all (returns 403 on POST /login) without an approved token. The bad thing about this change is that users may be confused when they see the 403 error screen, but it does solve the problem of the client spamming the server with login attempts.

This also adds a system to allow users to "approve" an IP automatically. Once an IP is approved, they will be able to log into the game with no interaction on the website. They can, of course, revoke the IP from being automatically signed in.

I also touched up the authentication page. Here's how it looks:
image

I added a new page to manage the whitelisted IPs a user has:
image

In my testing, the game will still send some spam sometimes (especially after rejecting EULA after logging in) but it will only send 3 attempts. I will try to remedy this at a later date, but for now this is an improvement from the infinite spam we saw last time.

Closes #79

### DIFF
--- a/ProjectLighthouse/Pages/ExternalAuth/AuthenticationPage.cshtml
+++ b/ProjectLighthouse/Pages/ExternalAuth/AuthenticationPage.cshtml
@@ -14,8 +14,22 @@
 else
 {
     <p>You have @Model.AuthenticationAttempts.Count authentication attempts pending.</p>
+    @if (Model.IpAddress != null)
+    {
+        <p>This device's IP address is <b>@(Model.IpAddress.ToString())</b>. If this matches with an authentication attempt below, then it's safe to assume the authentication attempt came from the same network as this device.</p>
+    }
+
+    <a href="/authentication/autoApprovals">
+        <button class="ui small blue button">
+            <i class="cog icon"></i>
+            <span>Manage automatically approved IP addresses</span>
+        </button>
+    </a>
     <a href="/authentication/denyAll">
-        <button class="ui small red button">Deny all</button>
+        <button class="ui small red button">
+            <i class="x icon"></i>
+            <span>Deny all</span>
+        </button>
     </a>
 }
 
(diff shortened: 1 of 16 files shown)


ProjectLighthouse/Pages/ExternalAuth/AuthenticationPage.cshtml
```diff
@@ -14,22 +14,48 @@
...
+        <i class="x icon"></i>
+        <span>Deny all</span>
+    </button>
+</a>
+
 @foreach (AuthenticationAttempt authAttempt in Model.AuthenticationAttempts)
 {
     DateTimeOffset timestamp = DateTimeOffset.FromUnixTimeSeconds(authAttempt.Timestamp);
     <div class="ui red segment">
         <p>A <b>@authAttempt.Platform</b> authentication request was logged at <b>@timestamp.ToString("MM/dd/yyyy @ h:mm tt") UTC</b> from the IP address <b>@authAttempt.IPAddress</b>.</p>
```
LumaLivy, line 40: I know this isn't really something new, but I'm wondering if we should maybe move the `MM/dd/yyyy @ h:mm tt` date string into a title attribute so the user can see the absolute time when hovering, and then have something more human readable for glancing:

(i.e. the user sees "A request was logged **a few seconds ago** from the IP address [...]" etc.

I know there are JS libraries that do this relative time wording like  [momentjs](https://momentjs.com/docs/#/customization/relative-time/), but there may be something similar that's better suited for this project.

Furthermore, we might want to have a little [!] warning indicator somewhere here if the IP address doesn't match the one shown up top, just to bring attention to the fact that the request may have come from outside their network. Geting even fancier, maybe we can include the approximate region/country of the request? Not sure what it takes to get that kind of info, but it definitely helps someone put more pieces of the puzzle together if they aren't familiar with IP addresses.

LumaLivy (approved):
Looks good to me. I commented above with some suggestions for the auth page, but they're mostly opinionated/QoL stuff.

jvyden:
Yeah, that will probably have to be done in a separate MR. I wanted to use a relative date system such as the js one you mentioned everywhere, so I'll probably do it then.

PR #106, opened 2021-12-14
Add more slot categories

LumaLivy:
## Additions
This PR adds most of the missing search pages from the LBP2 community page (most hearted, highest rated, etc) with support for each of the timespans (weekly, monthly, all-time)

## Changes
Cool levels now grabs the highest rated levels of the week (which is a bit more accurate of a placeholder to the original purpose of cool pages than using lucky dip). We can make a better/separate algo for cool levels when we get recent activity going.

## What needs work
I wanted to start supporting some of the query parameters for filtering, so I had a look at filtering by game. The `gameFilterType` query parameter indicates "both" in LBP2 when the game wants LBP1+2 levels, but, I'm not sure what LBP3 uses when searching for results from all/multiple games (or if it uses different endpoints entirely. I don't have a way to test it myself so functionality for LBP3 searching for multiple `GameVersion`s wasn't included). This also includes move pack levels (separate checkbox from LBP2) and vita/psp, as the other unknown/probably separate `GameVersion`s.

Edit: referencing #31 , this adds most of the categories minus recently played and busiest.


jvyden (changes requested):
Not a fan of how duplicated/copy-pasted all of the functions are. That can definitely be reduced by moving some of that to static methods in a helper.

jvyden (approved):
I did it myself :trollface:

PR #110, opened 2021-12-18
(Website) View location of photo subjects in photo

### LINKED ISSUE #91: Show *where* people are in uploaded photo on website
The website already says the people in an uploaded photo that has people in it but it would also be cool to see where the people are implemented.

LumaLivy:
This is a feature proposal that could possibly close #91 if approved.

I have implemented a simple JS+HTML Canvas username label+boundary renderer to allow anyone to view the location of photo subjects within a photo, through the main website.

[demo gif](https://media.discordapp.net/attachments/915460867506712637/921616288005120100/Peek_2021-12-17_23-02.gif?width=1067&height=686)

Subject bounds are hidden by default, requiring the user to hover over the subject list to view them. Furthermore, labels will move to stay within the frame if they were to render outside the canvas.

Here is a quick screenshot showing off the colour (multiple players) and label shifting functionality.

image

### Notes
- This feature works site-wide, as long as the photo is rendered as a `PhotoPartial`. Labels and their text will maintain their absolute scale on the page (so that it's still possible to read the label when there's a small `PhotoPartial`
- The labels will not shift to avoid each other -- overlap is possible if many players are packed close
  - this is due to the fact that the labels are all drawn to the same canvas. It's possible to draw a canvas for each label but it may lead to performance issues or require a rework of logic. Coding collision logic between labels seemed too out of scope for the scale of this feature, but maybe we'll see?

TLDR: This approach adds a canvas overlay to each `PhotoPartial` that renders subjects and their locations using the HTML canvas.

### DIFF
--- a/ProjectLighthouse/Pages/Layouts/BaseLayout.cshtml
+++ b/ProjectLighthouse/Pages/Layouts/BaseLayout.cshtml
@@ -64,6 +64,17 @@
             gtag('config', '@ServerSettings.Instance.GoogleAnalyticsId');
         </script>
     }
+
+    <style>
+        canvas.photo-subjects {
+            opacity: 1;
+            transition: opacity 0.3s;
+        }
+
+        canvas.hide-subjects {
+            opacity: 0;
+        }
+    </style>
 </head>
 <body>
 <div class="pageContainer">
--- a/ProjectLighthouse/Pages/Layouts/BaseLayout.cshtml
+++ b/ProjectLighthouse/Pages/Layouts/BaseLayout.cshtml
@@ -64,17 +64,6 @@
             gtag('config', '@ServerSettings.Instance.GoogleAnalyticsId');
         </script>
     }
-
-    <style>
-        canvas.photo-subjects {
-            opacity: 1;
-            transition: opacity 0.3s;
-        }
-
-        canvas.hide-subjects {
-            opacity: 0;
-        }
-    </style>
 </head>
 <body>
 <div class="pageContainer">
--- a/ProjectLighthouse/Pages/Partials/PhotoPartial.cshtml
+++ b/ProjectLighthouse/Pages/Partials/PhotoPartial.cshtml
@@ -33,9 +33,8 @@
 
 
 <script>
-    // set timeout so that the image heights have been calculated
-    // (after document renders)
-    window.setTimeout(function () {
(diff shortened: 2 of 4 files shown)


ProjectLighthouse/Pages/Layouts/BaseLayout.cshtml
```diff
@@ -64,6 +64,17 @@
...
+    <style>
+        canvas.photo-subjects {
+            opacity: 1;
+            transition: opacity 0.3s;
+        }
+
+        canvas.hide-subjects {
+            opacity: 0;
+        }
+    </style>
```
jvyden, line 77: I'd move this to its own CSS file.

ProjectLighthouse/Pages/Partials/PhotoPartial.cshtml
```diff
@@ -16,7 +24,103 @@
...
+        canvas.height = image.clientHeight; // space for names to hang off
+
+        const w = canvas.width;
+        const h = canvas.height;
+
+        // halfwidth, halfheight
+        const hw = w / 2;
+        const hh = h / 2;
+
+        const colours = ["#a5599f", "#477f84", "#5fa559", "#a5595f"];
```
jvyden, line 65: It would be cool to have these be popit colors:
```suggestion
        const colours = ["#96dd3c", "#ceb424", "#cc0a1d", "#c800cc"];
```

ProjectLighthouse/Pages/Partials/PhotoPartial.cshtml
```diff
@@ -16,7 +24,103 @@
...
+            context.lineWidth = 3;
+            context.strokeStyle = colour;
+            context.rect(bx, by, bw, bh);
+            context.stroke();
+
+            // Move into relative coordinates from bounding box
+            context.translate(bx, by);
+
+            // Username label
+            context.font = "16px Arial";
```
jvyden, line 91: I'd use Fomantic's font:
```suggestion
            context.font = "16px Lato";
```

jvyden (changes requested):

jvyden:
Consider the above just "comment" instead of "request changes", apart from the first comment its just some things that should be considered.

jvyden:
lmao screw it ill just do it myself :sunglasses:

PR #132, opened 2022-01-31
Added basic search functionality and fixed some bugs

Slendy:
This PR adds basic search functionality to users, photos, and levels as well as fixes some bugs like:
* when there are zero photos/slots/users the invalid page redirect will cause an error when trying to clamp because the page number will be -1
* Even if webhook integration is disabled, the WebhookHelper will still try to initialize a WebhookClient which will cause an error if there is no URL specified.
* If you have never run the instance before, the png folder won't get created and as a result, all related features fail to work
* Fixed a visual issue with the login and register page where after you hit submit the password field gets updated to the hashed version which causes the user to see the login field filled with characters. My fix still hashes the password and doesn't send the original password but now the password remains visually after hitting submit
# Before
Before
# After
Before

### DIFF
--- a/ProjectLighthouse/Controllers/ResourcesController.cs
+++ b/ProjectLighthouse/Controllers/ResourcesController.cs
@@ -57,10 +57,6 @@ public class ResourcesController : ControllerBase
     {
         string path = $"png/{hash}.png";
 
-        if (!IOFile.Exists("png"))
-        {
-             FileHelper.EnsureDirectoryCreated(Path.Combine(Environment.CurrentDirectory, "png"));
-        }
         if (IOFile.Exists(path))
         {
             return this.File(IOFile.OpenRead(path), "image/png");
--- a/ProjectLighthouse/Controllers/ResourcesController.cs
+++ b/ProjectLighthouse/Controllers/ResourcesController.cs
@@ -1,4 +1,5 @@
 #nullable enable
+using System;
 using System.IO;
 using System.Linq;
 using System.Threading.Tasks;
@@ -56,6 +57,10 @@ public class ResourcesController : ControllerBase
     {
         string path = $"png/{hash}.png";
 
+        if (!IOFile.Exists("png"))
+        {
+             FileHelper.EnsureDirectoryCreated(Path.Combine(Environment.CurrentDirectory, "png"));
+        }
         if (IOFile.Exists(path))
         {
             return this.File(IOFile.OpenRead(path), "image/png");
--- a/ProjectLighthouse/Pages/LoginForm.cshtml
+++ b/ProjectLighthouse/Pages/LoginForm.cshtml
@@ -14,7 +14,7 @@
         const passwordInput = document.getElementById("password");
         const passwordSubmit = document.getElementById("password-submit");
         
-        passwordSubmit.value = sha256(passwordInput);
+        passwordSubmit.value = sha256(passwordInput.value);
         
         return true;
     }
--- a/ProjectLighthouse/Pages/LoginForm.cshtml
+++ b/ProjectLighthouse/Pages/LoginForm.cshtml
@@ -11,9 +11,10 @@
 
 <script>
     function onSubmit(form) {
-        const passwordHashed = document.getElementById("password");
-        const passwordOriginal = document.getElementById("password-original");
-        passwordHashed.value = sha256(passwordOriginal.value);
+        const passwordInput = document.getElementById("password");
+        const passwordSubmit = document.getElementById("password-submit");
+        
+        passwordSubmit.value = sha256(passwordInput);
         
         return true;
     }
@@ -43,8 +44,8 @@
     <div class="field">
         <label>Password</label>
         <div class="ui left icon input">
-            <input type="password" id="password-original" placeholder="Password">
-            <input type="hidden" name="password" id="password">
+            <input type="password" id="password" placeholder="Password">
+            <input type="hidden" id="password-submit" name="password">
             <i class="lock icon"></i>
         </div>
     </div>
--- a/ProjectLighthouse/Pages/LoginForm.cshtml
+++ b/ProjectLighthouse/Pages/LoginForm.cshtml
@@ -11,9 +11,9 @@
 
 <script>
     function onSubmit(form) {
-        const password = form['password'];
-        
-        password.value = sha256(password.value);
+        const passwordHashed = document.getElementById("password");
+        const passwordOriginal = document.getElementById("password-original");
+        passwordHashed.value = sha256(passwordOriginal.value);
         
         return true;
     }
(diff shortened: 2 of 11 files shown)


ProjectLighthouse/Controllers/ResourcesController.cs
```diff
@@ -56,6 +57,10 @@ public IActionResult GetGameImage(string hash)
     {
         string path = $"png/{hash}.png";
 
+        if (!IOFile.Exists("png"))
+        {
+             FileHelper.EnsureDirectoryCreated(Path.Combine(Environment.CurrentDirectory, "png"));
+        }
```
jvyden, line 63: I would put this right before images get converted in `Program`.

ProjectLighthouse/Pages/SlotsPage.cshtml.cs
```diff
@@ -22,20 +22,26 @@ public class SlotsPage : BaseLayout
...
 
     public List<Slot> Slots;
+
+    public string SearchValue;
+
     public SlotsPage([NotNull] Database database) : base(database)
     {}
 
-    public async Task<IActionResult> OnGet([FromRoute] int pageNumber)
+    public async Task<IActionResult> OnGet([FromRoute] int pageNumber, [FromQuery] string name)
```
jvyden, line 31: I'd mark this as nullable.

ProjectLighthouse/Pages/UsersPage.cshtml.cs
```diff
@@ -22,20 +21,24 @@ public class UsersPage : BaseLayout
 
     public List<User> Users;
 
+    public string SearchValue;
+
     public UsersPage([NotNull] Database database) : base(database)
     {}
 
-    public async Task<IActionResult> OnGet([FromRoute] int pageNumber)
+    public async Task<IActionResult> OnGet([FromRoute] int pageNumber, [FromQuery] string name)
```
jvyden, line 29: I'd mark this as nullable.

ProjectLighthouse/Pages/RegisterForm.cshtml
```diff
@@ -10,11 +10,12 @@
...
-        const confirmPassword = form['confirmPassword'];
-        
-        password.value = sha256(password.value);
-        confirmPassword.value = sha256(confirmPassword.value);
+        const passwordInput = document.getElementById("password");
+        const confirmPasswordInput = document.getElementById("confirmPassword");
+        const password = document.getElementById("password-submit");
+        const confirmPassword = document.getElementById("confirm-submit");
+        password.value = sha256(passwordInput.value);
+        confirmPassword.value = sha256(confirmPasswordInput.value);
```
jvyden, line 18:
```suggestion
        const passwordSubmit = document.getElementById("password-submit");
        const confirmPasswordSubmit = document.getElementById("confirm-submit");
        
        passwordSubmit.value = sha256(passwordInput.value);
        confirmPasswordSubmit.value = sha256(confirmPasswordInput.value);
```

ProjectLighthouse/Pages/LoginForm.cshtml
```diff
@@ -43,7 +43,8 @@
     <div class="field">
         <label>Password</label>
         <div class="ui left icon input">
-            <input type="password" name="password" id="password" placeholder="Password">
+            <input type="password" id="password-original" placeholder="Password">
+            <input type="hidden" name="password" id="password">
```
jvyden, line 47:
```suggestion
            <input type="password" id="password" placeholder="Password">
            <input type="hidden" name="password-submit" id="password">
```

ProjectLighthouse/Pages/LoginForm.cshtml
```diff
@@ -11,9 +11,9 @@
 
 <script>
     function onSubmit(form) {
-        const password = form['password'];
-        
-        password.value = sha256(password.value);
+        const passwordHashed = document.getElementById("password");
+        const passwordOriginal = document.getElementById("password-original");
+        passwordHashed.value = sha256(passwordOriginal.value);
```
jvyden, line 16: The naming is a little inconsistent with `RegisterForm`, how about this?

```suggestion
        const passwordInput = document.getElementById("password");
        const passwordSubmit = document.getElementById("password-submit");
        
        passwordSubmit.value = sha256(passwordInput.value);
```

jvyden (changes requested):
Thanks for the quick work and additional bug-fixes!

Just a couple of small things:

ProjectLighthouse/Pages/SlotsPage.cshtml.cs
Slendy: Do you mean adding `[CanBeNull]`?

jvyden: That would also work, but I strongly prefer using the syntax C# comes with. You can mark it by adding `?` to the end of a type, so here it would be `[FromQuery] string? name`.

jvyden:
Hmm, web tests are failing. Perhaps because the password inputs were renamed?

ProjectLighthouse/Pages/SlotsPage.cshtml.cs
Slendy: I tried that but the IDE was correcting it to the other form so I had to go and disable it

jvyden: Yeah, Rider (I assume that's what you're using) likes to do that. How did you disable that?

Slendy: Settings -> Editor -> General -> Typing assistance then scroll down to C# and turn off annotate nullability

Slendy:
The form should still be submitting the same data for login and registration

ProjectLighthouse/Pages/SlotsPage.cshtml.cs
jvyden: Gotcha. Thanks for the tip!

jvyden:
The web tests are actually interacting with the page by firing up a chromium tab and programatically typing into the fields.

[Here's the relevant test](https://github.com/LBPUnion/ProjectLighthouse/blob/main/ProjectLighthouse.Tests.WebsiteTests/AuthenticationTests.cs).

I suspect it's just not able to find the field anymore so the login test fails.

jvyden (approved):
Code-side of things look good now! I'll approve now, but tests do need to pass before I actually merge.

jvyden:
Just a heads up regarding EntityFramework:
Calling `.Include` in a query will give the resulting object access to a foreign key. Not calling it will result in the foreign key being null. (a foreign key is more or less a pointer in the database that points to a row in another table)

So, if I did `Photo p = this.Database.Photos.Include(p => p.Creator).First()`, `p` will have access to the `Creator` object. If I don't call `.Include` (`Photo p = this.Database.Photos.First()`) `Creator` will be null.

When you made your change to add a search box to `PhotosPage` you removed the include call, thus resulting in the creator's name being blank:
image

Slendy:
Yeah I see, I assumed the Include was just there to fetch all objects because there was no query.

Slendy:
I know why not a single test has worked so far, so basically on the first commit, the test failed because the expected login id changed. When I changed the naming scheme though I left out the call to get the value of the password field. All tests pass now though

jvyden:
Ah, gotcha. That would do it, lol.

This should merge once tests pass now.

PR #145, opened 2022-02-05
Refactor comments and other features

Slendy:
- Added Level comments and the ability to yay or boo all comments
- You can view level comments on the website as well
- Added proper comment deletion
- Deleted comments on the website appear the same as they do in-game
- Added the config option to disable reviews, level comments, and user comments
- Fixed the visual hash bug on the reset password page
- Fixed author labels not showing up in-game in certain scenarios

What still needs to be improved/added:
- The ability to post level comments and user comments from the website
- The ability to see comment ratings and boo or yay the comment

Before label fix:
Before

After label fix:
After


Slendy:
@jvyden I think I messed up the migration because the tests failing is related to MySQL itself. Could you take a look?

Slendy:
I guess the EF migration didn't add the correct annotations to the migration file so it wasn't being recognized and wouldn't get applied. No idea why that happened

ProjectLighthouse/Migrations/20220205132152_CommentRefactor.cs
```diff
@@ -0,0 +1,114 @@
...
+                table: "Comments");
+
+            migrationBuilder.DropIndex(
+                name: "IX_Comments_TargetUserId",
+                table: "Comments");
+
+            migrationBuilder.RenameColumn(
+                name: "TargetUserId",
+                table: "Comments",
+                newName: "Type");
```
jvyden, line 27: Does this migrate existing comments properly? This doesn't seem right...

ProjectLighthouse/Pages/SlotPage.cshtml
```diff
@@ -54,6 +57,36 @@
...
+    @if (Model.Comments.Count == 0)
+    {
+        <p>There are no comments.</p>
+    }
+
+    @foreach (Comment comment in Model.Comments!)
+    {
+        DateTimeOffset timestamp = DateTimeOffset.FromUnixTimeSeconds(comment.Timestamp / 1000);
+        StringWriter messageWriter = new();
+        HttpUtility.HtmlDecode(comment.getComment(), messageWriter);
```
jvyden, line 71: This doesnt allow XSS does it?

jvyden (changes requested):
Nice work!
As always, a few concerns...

jvyden:
> I guess the EF migration didn't add the correct annotations to the migration file so it wasn't being recognized and wouldn't get applied. No idea why that happened

Perhaps the designer.cs file with the annotations (which we end up removing) didnt get generated? Not sure on that one, either.

ProjectLighthouse/Pages/SlotPage.cshtml
Slendy: It shouldn't, it's straight from the user class
test

ProjectLighthouse/Migrations/20220205132152_CommentRefactor.cs
Slendy: No it doesn't, good catch.

jvyden (approved):
Just tested it out on my local dev instance, works well!

Thanks for the contribution!

PR #167, opened 2022-02-14
Added more missing Title IDs

FridiNaTor1:
Added a few Title IDs including LBP 1, 2 and Vita missing from the list.


turecross321:
?

jvyden (approved):

PR #180, opened 2022-02-21
Fix review button not showing up

### LINKED ISSUE #107: Create Review button does not appear ingame
Reviews are now implemented in the game as of #102, which closes #27, but creating reviews is a bit scuffed. 

To allow review uploads, I made a workaround that involves having the server send an empty "fake" review to the player upon playing a level, so that they may edit the review and force the review to upload (the same endpoint is called for editing a review/uploading one). This way, only the player can see their empty "fake" review and no one else can, until the review is uploaded. The fake review is not stored in the database until it is uploaded as a real review either. It's basically a convoluted way of allowing the user to open the review editor in-game.

Ideally, a way is found to re-enable the "Leave a Review!" button in-game, which will prompt the review editor to open without having to edit an existing review. The button may be hidden due to a missing network configuration option, or missing data in the XML schema, but the true reason is not yet known.

Slendy:
Review button go brr

Closes #107


jvyden (approved):
I still cant believe it was that obvious.

PR #189, opened 2022-02-22
Send yourReview with slot request

Slendy:
Including yourReview was already implemented but somebody forgot to actually fetch the user's potential review in SlotController
This now makes the game recognize that it has already posted a review and replaces the review button with your review at the top of the list

Before:
Before

After:
After


jvyden (approved):
Nice catch! I was wondering why the review process felt a little different from what I remembered.

PR #246, opened 2022-03-19
Clean up updateUser and some small changes

### LINKED ISSUE #168: Grief Reports fail to submit if it tries to upload an unsafe file
From what I can gather it will attempt to upload relevant files (meshes, models, and other unsafe files we have blocked) when submitting a grief report

So when the option to check for unsafe files is enabled it will attempt to upload all these unsafe files and the grief report will not submit. 

Not sure how I'm going to resolve this.

Slendy:
I don't know if the other games behave like this but in LBP2 when your biography is empty it shows "[name] hasn't introduced themselves yet" so there is no need for us to manually set it upon creation. This can also be seen as a small quality of life change because right now when you create a new profile you have to manually delete the starting biography text instead of it just being empty.

The resource controller returning 409 instead of 422 when it encounters an unsafe file should also fix #168 but I need someone to test because I am unable to replicate the issue.

### DIFF
--- a/ProjectLighthouse/Pages/UserPage.cshtml
+++ b/ProjectLighthouse/Pages/UserPage.cshtml
@@ -83,7 +83,14 @@
     <div class="eight wide column">
         <div class="ui blue segment">
             <h2>Biography</h2>
-            <p>@Model.ProfileUser.Biography</p>
+            @if (Model.ProfileUser.Biography.Length == 0)
+            {
+                <p>@Model.ProfileUser.Username hasn't introduced themselves yet</p>
+            }
+            else
+            {
+                <p>@Model.ProfileUser.Biography</p>
+            }
         </div>
     </div>
     <div class="eight wide column">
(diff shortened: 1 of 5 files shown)


ProjectLighthouse/Pages/UserPage.cshtml
```diff
@@ -83,7 +83,14 @@
     <div class="eight wide column">
         <div class="ui blue segment">
             <h2>Biography</h2>
-            <p>@Model.ProfileUser.Biography</p>
+            @if (Model.ProfileUser.Biography.Length == 0)
```
jvyden, line 86: I'd just do `string.IsNullOrEmpty` here.

jvyden (changes requested):
Just a small nitpick, otherwise code looks fine.

Testing grief reports on a PS3 as we speak.

jvyden (approved):
Grief reports work properly now. Nice job! Interesting how 409 is a success to LBP :thinking:

PR #248, opened 2022-03-21
Improve user profile picture loading speeds

Slendy:
The /users endpoint in-game actually only cares about the user's icon so there is no point in serializing their entire profile. I also used projection so only the icon and username column are fetched which in the worst-case sped up performance by about half compared to just fetching the user normally.

From my testing, the current implementation could take up to half a second for just 3 profiles that weren't fully filled out so I imagine this could take a long time for larger queries. The new method takes just 100ms in the same worst-case scenario. In the average scenario, the speedup is about 8x going from 80ms on average to less than 10ms on average.

The effects of this can be seen everywhere because loading levels, comments, reviews, scores, and practically anything user-related involves loading their profile picture.

### DIFF
--- a/ProjectLighthouse/Controllers/GameApi/UserController.cs
+++ b/ProjectLighthouse/Controllers/GameApi/UserController.cs
@@ -41,7 +41,7 @@ public class UserController : ControllerBase
                 u.Username,
                 u.IconHash,
             }).FirstOrDefaultAsync();
-        if (partialUser == null) return "";
+        if (partialUser == null) return null;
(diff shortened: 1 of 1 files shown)


jvyden (approved):
Yeah, I noticed the giant ?u queries took a while. It's definitely important for this to be a fast operation, considering that LBP1 requests profiles for all friends (which can go up to I believe 100?)

The implementation is definitely not one I would've thought of. I assumed most of the timing was due to getting counts/other information off the user but I guess not.

Good work!

PR #262, opened 2022-04-02
Add ability to grant a user additional slots

### LINKED ISSUE #232: Allow increasing slot limit for specific users
Shanzenos said this would be useful for publishing archived levels

jvyden:
This moves the ban user button to a new Admin Options panel at the bottom of a user page:
image

The value you set will be applied in addition to the instance's default slot quota, 
So, for example, if the server's slot quota was 50 and you granted 14 additional slots, the total slot quota for that user would be 64.

*Closes #232*

### DIFF
--- a/ProjectLighthouse/Pages/UserPage.cshtml
+++ b/ProjectLighthouse/Pages/UserPage.cshtml
@@ -73,12 +73,7 @@
             </a>
         }
         @if (Model.User != null && Model.User.IsAdmin && !Model.ProfileUser.Banned)
-        {
-            <a class="ui red button" href="/admin/user/@Model.ProfileUser.UserId/ban">
-                <i class="ban icon"></i>
-                <span>Ban User</span>
-            </a>
-        }
+        {}
     </div>
     <div class="eight wide column">
         <div class="ui blue segment">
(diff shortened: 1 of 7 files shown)


ProjectLighthouse/Pages/UserPage.cshtml
```diff
@@ -118,4 +113,32 @@
...
+        <form method="post" action="/admin/user/@Model.ProfileUser.UserId/setGrantedSlots">
+            @Html.AntiForgeryToken()
+            <div class="ui left action input">
+                <button type="submit" class="ui blue button">
+                    <i class="pencil icon"></i>
+                    <span>Set Granted Slots</span>
+                </button>
+                <input type="text" name="grantedSlotCount" placeholder="Granted Slots" value="@Model.ProfileUser.AdminGrantedSlots">
+            </div>
+        </form>
```
Slendy, line 141: Can this form be turned into a partial so there is no duplicated code between this and `AdminSetGrantedSlotsPage.cshtml`?

ProjectLighthouse/Pages/UserPage.cshtml
```diff
@@ -73,12 +73,7 @@
             </a>
         }
         @if (Model.User != null && Model.User.IsAdmin && !Model.ProfileUser.Banned)
```
Slendy, line 75: You can probably delete this and the empty braces below it
```suggestion
```

Slendy (changes requested):
Looks good.

Just two minor suggestions about duplicated code and an empty if statement.

ProjectLighthouse/Pages/UserPage.cshtml
jvyden: Probably. Will do so in the next commit.

ProjectLighthouse/Pages/UserPage.cshtml
```diff
@@ -72,7 +72,6 @@
                 <span>Reset Password</span>
             </a>
         }
-        @if (Model.User != null && Model.User.IsAdmin && !Model.ProfileUser.Banned)
         {}
```
Slendy, line 75: You forgot the curly braces
```suggestion
```

Slendy (approved):

ProjectLighthouse/Pages/UserPage.cshtml
jvyden: DARNIT

PR #263, opened 2022-04-03
Several improvements to Grief Reports

jvyden:
First and foremost, I made some improvements to the actual reports themselves.

I added some spacing to different areas of the information, moved the "hover here to view area" text to be the actual image, and increased the size of the buttons a tiny bit.

image

I also fixed the bug where the canvas would not display. The bug was the height was being set to 0 for some reason.

Lastly, I added the ability to "dismiss" a report instead of having the only available option be going hail mary and deleting all related assets. This simply removes the report from the database, nothing special.


jvyden:
That's not the intention behind the dismiss button, it's more of a "this report is invalid" or "we handled the report manually" button to remove the report from the menu without deleting all the assets as the other button does.

Slendy:
Yeah, I see for the other assets but the jpegHash is unique for each report so if you are removing the report then it doesn't make sense to keep it.

jvyden:
Ah, you mean the actual image sent in the report. I'll implement nuking that in the next commit.

jvyden:
Forgot to mention; good catch!

Slendy (approved):

PR #264, opened 2022-04-05
Add reviews to slot page

### LINKED ISSUE #234: Show reviews on the website

jvyden:
Here's what it looks like on a level with mixed reviews:
image

Vita's review icons are a little smaller for some strange reason, but they still look alright:
image

- [x] Depends on #267

*Closes #234*

### DIFF
--- a/ProjectLighthouse/Pages/SlotPage.cshtml
+++ b/ProjectLighthouse/Pages/SlotPage.cshtml
@@ -1,5 +1,7 @@
 @page "/slot/{id:int}"
 @using LBPUnion.ProjectLighthouse.Helpers.Extensions
+@using LBPUnion.ProjectLighthouse.Types.Reviews
+@using LBPUnion.ProjectLighthouse.Types.Settings
 @model LBPUnion.ProjectLighthouse.Pages.SlotPage
 
 @{
@@ -8,6 +10,8 @@
 
     Model.Title = Model.Slot.Name;
     Model.Description = Model.Slot.Description;
+
+    bool isMobile = this.Request.IsMobile();
 }
 
 @await Html.PartialAsync("Partials/SlotCardPartial", Model.Slot, new ViewDataDictionary(ViewData)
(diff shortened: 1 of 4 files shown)


ProjectLighthouse/Pages/SlotPage.cshtml
```diff
@@ -53,14 +57,84 @@
...
 
-@await Html.PartialAsync("Partials/CommentsPartial")
+            @foreach (Review review in Model.Reviews)
+            {
+                string faceHash = review.Thumb switch {
+                    -1 => review.Reviewer.BooHash,
+                    0 => review.Reviewer.MehHash,
+                    1 => review.Reviewer.YayHash,
+                    
+                    _ => throw new ArgumentOutOfRangeException(),
```
Slendy, line 90: The /postReview endpoint doesn't clamp the Thumb value so a user can make their profile cause an error. This should either have a default value or the Thumb value needs to properly get limited in ReviewController.

ProjectLighthouse/Pages/SlotPage.cshtml.cs
```diff
@@ -43,6 +45,19 @@ public async Task<IActionResult> OnGet([FromRoute] int id)
             this.Comments = new List<Comment>();
         }
 
+        if (this.ReviewsEnabled)
+        {
+            this.Reviews = await this.Database.Reviews.Include(r => r.Reviewer)
+                .OrderByDescending(r => r.Timestamp)
```
Slendy, line 51:
```suggestion
                .OrderByDescending(r => r.ThumbsUp - r.ThumbsDown)
                .ThenByDescending(r => r.Timestamp)
```
This should be ordered by either thumbsUp or the result of thumbs up minus thumbs down

Slendy (changes requested):
Nice work. I only noticed a few small things that could be improved. I also think in the future we should add pages to reviews, comments, and photos on user and slot pages to conserve space.

ProjectLighthouse/Pages/SlotPage.cshtml
jvyden: Good to know. Are there any other places you can think of where ratings aren't clamped?

Slendy: In my purge fixes branch I already fixed comment ratings not being clamped, but you could fix it here if you wanted

jvyden: Hmm. If that's on a separate commit I could cherry-pick it into master. If not I can fix it in this MR. Up to you.

Slendy: It'll be one big commit so I'll just do it in my PR.

jvyden: Cool. I'll wait until then.

Slendy (approved):

PR #265, opened 2022-04-05
Add some disclaimers to the README

jvyden:
The hope is that we can link to this section of the README whenever someone tries to ask for support for running an unofficial Project Lighthouse instance.

I also copypasted most of the legalties from the initial Lighthouse article and added it as a new section.


m88youngling (approved):
don't even worry about it

Issue #258, opened 2022-04-02
Slot descriptions are not truncated


jvyden:
This can cause Lighthouse to crash due to OOM because of multiple very large slot descriptions being loaded, e.g. when a user's earth that has many slots with long descriptions gets loaded

Slendy:
Closed in #267

PR #267, opened 2022-04-06
Fixes and improvements from Operation Purge

### LINKED ISSUE #261: XML Injection possible in level and profile descriptions
image
image
image
image

Slendy:
# Main fixes
* User biographies, reviews, and level names are now limited to 100 characters
* Level descriptions are now limited to 200 characters
* yourThumb is now properly sent with the review list
* Review ordering is now done by thumbs up minus thumbs down instead of just thumbs up
* Review thumbs are now recalculated every time someone changes their rating
* All XML that is deserialized is now sanitized by turning `<` and `>` into the safer `&lt;` and `&gt;`

I also added a maintenance job to clean up all existing unsafe XML that has been uploaded prior to this commit.

Closes #261 and #258


ProjectLighthouse/Helpers/ReflectionHelper.cs
```diff
@@ -0,0 +1,25 @@
+﻿#nullable enable
+using System.Reflection;
+
+namespace LBPUnion.ProjectLighthouse.Helpers;
+
+public static class ReflectionHelper
+{
+
+    public static void sanitizeStringsInClass(object? instance)
```
jvyden, line 9: Should be called `SanitizeStringsInClass`.

ProjectLighthouse/Helpers/ReflectionHelper.cs
```diff
@@ -0,0 +1,25 @@
...
+        PropertyInfo[] properties = instance.GetType().GetProperties();
+        foreach (PropertyInfo property in properties)
+        {
+            if (property.PropertyType != typeof(string)) continue;
+
+            string? before = (string?) property.GetValue(instance);
+            if (before == null) continue;
+            if (!before.Contains('>') || !before.Contains('<')) continue;
+            
+            property.SetValue(instance, before.Replace("<", "&lt;").Replace(">", "&gt;"));
```
jvyden, line 21: This could be written better for extensibility, say for example other characters can be abused in such a way, or we could even use this for censoring user input.

Here's what I'm thinking:
1. Have a dictionary mapping `char`s to `string`s, populated with the appropriate replacements:
```csharp
private static readonly Dictionary<char, string> charsToReplace = new() {
    {'<', "&lt;"},
    {'>', "&gt;"},
};
```
Probably not the right syntax but the idea is there.

2. Replace the check on line 19 with something like this:
```csharp
if(!charsToReplace.Any(before.Contains)) continue; // Stop if there are no characters that need to be replaced
```

3. Then, replace the characters like so:
```csharp
string after = "";

foreach(KeyValuePair<char, string> entry in charsToReplace)
{
    after = after.Replace(entry.Key, entry.Value);
}

property.SetValue(instance, after);
```

ProjectLighthouse/Maintenance/MaintenanceJobs/CleanupXMLInjection.cs
```diff
@@ -0,0 +1,57 @@
...
+
+        foreach (Photo photo in this.database.Photos)
+        {
+            ReflectionHelper.sanitizeStringsInClass(photo);
+        }
+
+        foreach (GriefReport report in this.database.Reports)
+        {
+            ReflectionHelper.sanitizeStringsInClass(report);
+        }
```
jvyden, line 53: Nit: This might look way nicer if we didn't use brackets here.

ProjectLighthouse/Maintenance/MaintenanceJobs/CleanupXMLInjection.cs
```diff
@@ -0,0 +1,57 @@
...
+using LBPUnion.ProjectLighthouse.Types.Profiles;
+using LBPUnion.ProjectLighthouse.Types.Reports;
+using LBPUnion.ProjectLighthouse.Types.Reviews;
+
+namespace LBPUnion.ProjectLighthouse.Maintenance.MaintenanceJobs;
+
+public class CleanupXmlInjection : IMaintenanceJob
+{
+    private readonly Database database = new();
+    public string Name() => "Cleanup unsanitized XML";
```
jvyden, line 15:
```suggestion
    public string Name() => "Sanitize user content";
```

ProjectLighthouse/Maintenance/MaintenanceJobs/CleanupXMLInjection.cs
```diff
@@ -0,0 +1,57 @@
...
+    public string Description() => "Sanitizes all strings in levels, reviews, comments, users, and scores";
```
jvyden, line 16:
```suggestion
    public string Description() => "Sanitizes all user-generated strings in levels, reviews, comments, users, and scores to prevent XML injection. Only needs to be run once.";
```

ProjectLighthouse/Pages/SlotPage.cshtml
```diff
@@ -31,7 +31,7 @@
     <div class="eight wide column">
         <div class="ui blue segment">
             <h2>Description</h2>
-            <p>@(string.IsNullOrEmpty(Model.Slot.Description) ? "This level has no description." : Model.Slot.Description)</p>
+            <p>@Html.Raw(string.IsNullOrEmpty(Model.Slot.Description) ? "This level has no description." : Model.Slot.Description)</p>
```
jvyden, line 34: The usage of `Html.Raw` on user input here makes me a little uncomfortable... 

Is this really the best way to show this?

ProjectLighthouse/Pages/UserPage.cshtml
```diff
@@ -82,7 +82,7 @@
             }
             else
             {
-                <p>@Model.ProfileUser.Biography</p>
+                <p>@Html.Raw(Model.ProfileUser.Biography)</p>
```
jvyden, line 85: See my `Html.Raw` comment

ProjectLighthouse/Pages/UserPage.cshtml
```diff
@@ -120,14 +120,14 @@
...
-            </a>
-        
-        <br><br>
+            <div>
+                <a class="ui red button" href="/admin/user/@Model.ProfileUser.UserId/ban">
+                    <i class="ban icon"></i>
+                    <span>Ban User</span>
+                </a>
+            </div>
+            <div class="ui fitted hidden divider"></div>
```
jvyden, line 129: What does this look like?

ProjectLighthouse/Helpers/ReflectionHelper.cs
```diff
@@ -0,0 +1,25 @@
+﻿#nullable enable
+using System.Reflection;
+
+namespace LBPUnion.ProjectLighthouse.Helpers;
+
+public static class ReflectionHelper
```
jvyden, line 6: This class could be merged with `CensorHelper`, something like `SanitizationHelper`?

jvyden (changes requested):
Nice work! This seems like it'll work pretty well, but I do see a way we can improve this in case we want to build on this further...

ProjectLighthouse/Pages/UserPage.cshtml
Slendy: image

ProjectLighthouse/Pages/SlotPage.cshtml
Slendy: User input should always get sanitized before reaching that but I'll change it to `HttpUtility.HtmlDecode` to be safe

jvyden (approved):

PR #285, opened 2022-04-14
Fix include slot location and creator arg

Slendy:
Someone kinda broke production but this should fix it.


jvyden:
Very busy right now, so we'll do it like this.

..Does it work?

Slendy:
It should, it can't be any worse than it is right now. Trying to load any levels is kinda broke in terms of creator and location.

jvyden (approved):
Works for me.

PR #307, opened 2022-05-14
Advanced level searches and visual fixes

Slendy:
This PR is primarily visual changes and fixes many dividers on the slot page that should or should not be there, a few examples are below. This also adds the ability to filter levels using the `by:` and `game:` keywords (similar to Discord) to filter by author and game version. Lastly, this fixes review special characters not being encoded like apostrophes.

<sub>(Search example)</sub>

# Comments
### Before
Comments didn't have a divider between the first comment and whatever was above it

### After
The divider is now placed depending on whether there are > 0 comments

# Reviews
### Before
The review list would show a divider regardless of whether there were any reviews

### After
Similarly to comments, the divider is only added if there are > 0 reviews


jvyden:
I hope I didn't just immediately cause a massive merge conflict...

jvyden (approved):
Looks good to me! In the future this could probably use some friendlier UI but for now this is functional.

Slendy:
I accidentally broke the total level count, I'm fixing it rn

jvyden:
Alright; I will merge that when that's ready.

PR #310, opened 2022-05-17
workaround for NETSDK1152 error

metraberryy:
gets rid of the "Found multiple publish output files with the same relative path" error i encountered while publishing from dotnet cli


jvyden (approved):
Publishing isn't really a part of the intended workflow but I'll merge anyways.

PR #351, opened 2022-07-15
Added user invite system

Zaprit:
Allow users to be invited via an api with a key for ease of user creation (on the request of @m88youngling )


Zaprit:
Upon more code writing the DB migrations on this still don't work 😩

Slendy:
I also had trouble doing migrations when I first started but I've learned that you have to copy the class attribute from the designer file that gets generated to the migration file. It looks something like this:
```
[DbContext(typeof(Database))]
[Migration("20220624210701_AddedPasswordResetTokens")]
```
Without the attribute then EntityFramework doesn't see and register it as an actual migration.

Zaprit:
> I also had trouble doing migrations when I first started but I've learned that you have to copy the class attribute from the designer file that gets generated to the migration file. It looks something like this:
> 
> ```
> [DbContext(typeof(Database))]
> [Migration("20220624210701_AddedPasswordResetTokens")]
> ```
> 
> Without the attribute then EntityFramework doesn't see and register it as an actual migration.

YUP, just figured that one out. so many hours wasted

ProjectLighthouse/PlayerData/APIKey.cs
```diff
@@ -0,0 +1,21 @@
+﻿
+using System;
+using System.ComponentModel.DataAnnotations;
+
+namespace LBPUnion.ProjectLighthouse.PlayerData
+{
+    public class APIKey
```
jvyden, line 7: Its worth noting that it would probably be helpful to link this key to a particular user for audit logging.

jvyden (changes requested):
Hello, thank you for your continued work on Lighthouse!

These are mostly nitpicks and grammar as usual, but I do have some comments on how the keys themselves are meant to work.

ProjectLighthouse.Servers.API/Controllers/UserEndpoints.cs
```diff
@@ -54,4 +55,34 @@ public IActionResult GetUserStatus(int id)
...
+        {
+
+            if (!string.IsNullOrWhiteSpace(this.Request.Headers["Authorization"]))
+            {
+                var AuthHeader = this.Request.Headers["Authorization"].ToString().Substring(6);
+
+                APIKey? apiKey = await this.database.APIKeys.FirstOrDefaultAsync(k => k.Key == AuthHeader);
+                if (apiKey == null) return this.Forbid();
+
+                if (!apiKey.Enabled) return this.Forbid();
```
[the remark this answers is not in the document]
jvyden: ok this looks stupid lol

ProjectLighthouse/PlayerData/APIKey.cs
Zaprit: So like a created by?

ProjectLighthouse/PlayerData/APIKey.cs
```diff
@@ -0,0 +1,21 @@
...
+        [Key]
+        public int Id { get; set; }
+
+        public string Description { get; set; }
+
+        public string Key { get; set; }
+
+        public DateTime Created { get; set; }
+
+        public bool Enabled { get; set; }
```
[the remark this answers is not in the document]
Zaprit: Yeah, i ended up at this conclusion myself, I think i didn't get round to pushing it though

ProjectLighthouse/PlayerData/APIKey.cs
jvyden: Yeah, exactly.

ProjectLighthouse/PlayerData/APIKey.cs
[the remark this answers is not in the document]
Zaprit: Ok, done on latest commits

ProjectLighthouse/PlayerData/APIKey.cs
Zaprit: while I agree, I think that's out of the scope of this PR, I had a bash at implementing that and it's a lot of changes to the command system.

jvyden (approved):
Code looks good to me now, going to test shortly.

jvyden:
Oops, build error :p

Zaprit:
> Oops, build error :p

Oof, forgot a reference to APIKey.Enabled

ProjectLighthouse.Servers.API/Controllers/UserEndpoints.cs
```diff
@@ -54,4 +55,32 @@ public IActionResult GetUserStatus(int id)
...
+    [HttpPost("user/inviteToken")]
+    public async Task<IActionResult> CreateUserInviteToken()
+    {
+        if (Configuration.ServerConfiguration.Instance.Authentication.PrivateRegistration ||
+            Configuration.ServerConfiguration.Instance.Authentication.RegistrationEnabled)
+        {
+
+            if (!string.IsNullOrWhiteSpace(this.Request.Headers["Authorization"]))
+            {
+                string authHeader = this.Request.Headers["Authorization"].ToString().Substring(6);
```
jvyden, line 68: Why are we using `Substring` here? It just made it confusing to send the request on my end.

jvyden:
Hmm, the registration token doesn't seem to get removed after using it...

jvyden:
> Hmm, the registration token doesn't seem to get removed after using it...

False alarm, it was just a token I previously created earlier and forgot about.

jvyden:
Alright, I don't see any more problems with this, merging...

PR #371, opened 2022-07-25
Added login redirection, level icon background and fixed bugs

### LINKED ISSUE #195: Add backdrop for transparent level and profile badges
There should be a dark gray background behind the level and profile icons so transparent graphics appear less odd

### LINKED ISSUE #280: User should be redirected to previous page after login
Currently, if you are on a page and try to log in, you land on the home page even if you were already on another page. A small quality of life change would be redirecting the user to the last page they were on. This can be done by including a callback URL in the login request that is either embedded in the login link for every page or fetched using `document.referrer` within JS.

### LINKED ISSUE #331: Round LBP1 star rating on website
floating point go brr
image
image

### LINKED ISSUE #350: There is a button to upvote your own comments, despite the fact that the server won't let you
image

Slendy:
This PR adds a redirect so that after a user logs in they are sent to the previous page they were on. (Closes #280)

This also fixes where if an image fails to load it displays a browser missing image icon. Now it will try and load the missing icon hash in its place. The onerror monstrosity required to do this could be moved into a separate function if it's that bad.

This PR also adds a background for transparent images which closes #195 
Before:

After:

Fixed comment vote buttons for your own comment. (closes #350)

Also closes #331


jvyden (approved):
LGTM, will look into test failures

jvyden:
image

hmmm, seems to have broken password input?

jvyden:
image

The code to generate the URL is stopping the code to generate the password because of this exception, causing the test to fail. It throws the exception because the document has no referrer. I'll push a commit to fix this in just a minute.

PR #372, opened 2022-07-25
Add support for ReCaptcha in addition to HCaptcha

jvyden:


ProjectLighthouse/Extensions/RequestExtensions.cs
```diff
@@ -15,7 +15,21 @@ namespace LBPUnion.ProjectLighthouse.Extensions;
 
 public static class RequestExtensions
 {
-
+    static RequestExtensions()
+    {
+        Uri captchaUri = ServerConfiguration.Instance.Captcha.Type switch
+        {
+            CaptchaType.HCaptcha => new Uri("https://hcaptcha.com"),
+            CaptchaType.ReCaptcha => new Uri("https://www.google.com/recaptcha/api/"),
```
Slendy, line 23: Does it matter that one has a trailing slash and the other doesn't?

Slendy (approved):
Looks good to me.

ProjectLighthouse/Extensions/RequestExtensions.cs
jvyden: In testing it does not.

Slendy: Should be fine then.

jvyden: Alright, merging then.

PR #385, opened 2022-07-30
Update default language to be en instead of en-US

### LINKED ISSUE #384: Language Assumption only detects en-US as English.
If you have multiple languages set in your browsers languages and en-US isn't amongst them or below en and en-gb won't get recognised, and instead the list proceeds until it finds a fitting one.

Mondanzo:
This allows browsers requesting over en-GB to still get the English version before another one.

Closes #384


jvyden (approved):
Looks good to me; was able to reproduce the issue and this solved it.

Thank you for your contribution to Project Lighthouse!

PR #386, opened 2022-07-31
Fix heart and queue lists

Slendy:
According to EntityFramework you have to call .Include before .Select otherwise the query fails.


jvyden (approved):
LGTM!

PR #387, opened 2022-07-31
Keep announce text blank and allow EULA to be customized a little further in terms of layout.

### LINKED ISSUE #376: /announce shouldn't return a line break if AnnounceText is blank

### LINKED ISSUE #380: Customization of EULA layout
The agpl license always comes first. I feel that the owner should be able to position it (but not remove it) a little further down to get some important points about rules and stuff out of the way, as most people would probably just read half of the agpl license and click accept.

Maybe do it with rext replacement like with announce, and if the %license text is not there, then throw a warning in console that exluding the license is not allowed and prepend it?

TorutheRedFox:
Changes message endpoints to make them more customizable, such as moving around the license to give a bit of important info before the license is given (such as where to find the rules governing the instance or whatever) and clearing out announce (as some people may find having to hit one extra button to get into the game a bit annoying ;P)

Doesn't allow people to remove license entirely from EULA though.

Resolves #380
Fixes #376

### DIFF
--- a/ProjectLighthouse.Servers.GameServer/Controllers/MessageController.cs
+++ b/ProjectLighthouse.Servers.GameServer/Controllers/MessageController.cs
@@ -84,7 +84,7 @@ along with this program.  If not, see <https://www.gnu.org/licenses/>.";
             $"token.ExpiresAt: {gameToken.ExpiresAt.ToString(CultureInfo.CurrentCulture)}\n" +
             "---DEBUG INFO---" +
             #endif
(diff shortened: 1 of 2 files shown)


jvyden:
Hello, thank you for your contribution to Project Lighthouse!

Unfortunately I can't accept this; we're planning for a different system as noted in #380.

You're welcome to create another PR for the fix for #376, though.

TorutheRedFox:
oh ok

PR #388, opened 2022-07-31
Don't send new line if announce text is blank

### LINKED ISSUE #376: /announce shouldn't return a line break if AnnounceText is blank

TorutheRedFox:
Reopening #387 without EULA changes
Fixes #376


jvyden (approved):
LGTM!

PR #393, opened 2022-08-02
Fix posters' comments not showing ratings

Slendy:
The last time I touched comments I accidentally made the comment ratings invisible 
image
This PR disables pointer events for the voting buttons instead of simply hiding them.


jvyden (approved):
LGTM

PR #395, opened 2022-08-02
Prevent grief report deletion from removing slot rootLevel

### LINKED ISSUE #353: The "Remove All Assets" button in the reports menu does not remove the level, it only breaks it
image

Slendy:
This kind of closes #353, I think it may also be a good idea to rename the remove all assets button because it insinuates that it will delete (unpublish) the level that the report was generated in.


jvyden:
Yeah, the way reports are handled kinda have to be reworked at some point; deleting everything in a report is kind of weird considering a report can just be about one resource in a sea of them.

jvyden (approved):
LGTM!

PR #400, opened 2022-08-04
Implement story mode player counts endpoint

Slendy:
I found some packet captures for /slots/developer, idk how the game actually uses this but I thought I may as well implement it.


TorutheRedFox:
isn't it to show current player counts in an infobubble above the level?

Slendy:
Individual story levels are supposed to already include the player counts though, so it would be redundant if it were just for displaying the info.

jvyden (approved):
LGTM

TorutheRedFox:
actually thinking about it, didn't developer slots show up in hearted lists? maybe it's for that?

PR #402, opened 2022-08-04
Include number of photos in slot serialization

Slendy:
These are necessary to hint to the game that the level has photos instead of the game having to manually check the /photos endpoint. This was also supposed to be in the original dev slots PR but I forgot about it.


jvyden (approved):
Oh. That would do it. I've been wondering why photos don't pop up in LBP2...

PR #440, opened 2022-08-19
Add support for moving levels with decoration tool

Slendy:
This PR implements support for moving levels when you are in decorate mode on your earth.


jvyden (approved):
Interesting, I thought this was already implemented last rework :thinking:

PR #463, opened 2022-08-31
Implement LBP1 tags, more strict resource checking, and more

### LINKED ISSUE #120: LBP1 Slot Categories
The following aren't implemented:
```
[HTTP] 404, 15ms: GET /LITTLEBIGPLANETPS3_XML/slots/highestRated?pageStart=1&pageSize=30
[HTTP] 404, 15ms: GET /LITTLEBIGPLANETPS3_XML/slots/tag?tag=TAG_Balancing&pageStart=1&pageSize=30
```

### LINKED ISSUE #237: Search results do not filter out uncompatiable levels
The search results do not filter out any levels. This for instance means that you'll see LBP2 levels while searching for levels on LBP1

### LINKED ISSUE #356: Level Tag Translation on website
There are many different tags that display something completely else on the website. Some examples are:

image
should be `Attract-o-Gel`

image
should be `Creatinator`

image
Controllinator?

### LINKED ISSUE #358: Show when someone was last online on website
it'd be cool if the website displayed when an offline user was last online
image

### LINKED ISSUE #436: Sub levels show in the search bar
**Describe the bug**
I was playing some levels but then I noticed that a sub level was showing up when I searched up a lebel

**To Reproduce**
1. Go to ‘LittleBigPlanet 2'
2. Click on 'Text Search'
3. Scroll down to 'Search up “CRLMSONFANG9'
4. See error There will be CrackedEdge & CrackedEdge (PART 2) which is a sub level 

**Expected behavior** 
The expected behavior would be that the (PART 2) level doesn’t show up

**Screenshots**
image
image

**Game(s) affected**
 - Platform Possibly all platforms
 - Game Possibly all the games

**Additional context**

### LINKED ISSUE #444: Tagging levels do not work on LBP 1
Tagging a level does not work on LBP 1, if I try Tagging a level nothing happens.

Slendy:
- Certain timestamps are now converted into the host machine's time zone
- Added user login and logout timestamps, when a user is offline last logout will be shown and if they have never logged in, it just shows 'Offline' (Closes #358)
- Fixed a bug where lbp1 play counts wouldn't be counted
- Added resource dependency table parsing which should help further with preventing modded content
- Fixed a bug where requests would still be processed even if they didn't include the content digest
- Fixed cases where certain requests would only include an X-Digest-B header and calculated the digest hash differently
- Implemented the ability to tag lbp1 levels (Closes #444)
- Photo resources are now deleted with the photo instead of staying on the server
- Labels are now validated when uploading levels and reviews
- Added more safety checks for uploading a score
- Fixed a bug where story mode levels would appear in a level search
- Fixed a bug where sub-levels would appear in a level search (Closes #436)
- Fixed a bug where levels from the wrong game would appear in a level search (Closes #237)
- Added label translation on the website (eg. DirectControl now appears as Controllinator) (Closes #356)
- Implemented the search tags and highest rated buttons on lbp1 (Closes #120)
- Implemented the find more like this button on lbp1
- Fixed a bug where /reviewsFor would cause an error if there were 0 reviews


ProjectLighthouse/Database.cs
```diff
@@ -451,6 +451,12 @@ public bool IsRegistrationTokenValid(string tokenString)
 
     public async Task RemoveExpiredTokens()
     {
+        foreach (GameToken token in this.GameTokens.Where(t => DateTime.Now > t.ExpiresAt).ToList())
+        {
+            User? user = await this.Users.FirstOrDefaultAsync(u => u.UserId == token.UserId);
+            if(user != null) user.LastLogout = TimeHelper.TimestampMillis;
+            this.GameTokens.Remove(token);
+        }
         this.GameTokens.RemoveWhere(t => DateTime.Now > t.ExpiresAt);
```
Zaprit, line 460: Is this necessary if on the above lines the gametokens are cleared out anyway

Zaprit (approved):
All seems good. Saw one bit of seemingly surplus code but then again might just be me not knowing what's going on

ProjectLighthouse/Database.cs
Slendy: Good catch, I forgot to get rid of the original line that removed game tokens

PR #491, opened 2022-09-23
Update README.md

Ezoiar:


m88youngling (approved):
When contributing, remember to add a comment explaining what you did.

README.md
```diff
@@ -71,7 +71,7 @@ information.
 |----------|------------------------|-------------------------------------------|-------------------------------|
 | LBP1     | Compatible             | Compatible                                | No next-gen equivalent        |
 | LBP2     | Compatible             | Compatible                                | No next-gen equivalent        |
-| LBP3     | Mostly compatible      | Mostly compatible                         | Incompatible                  |
+| LBP3     | Mostly compatible      | Compatible                         | Incompatible                  |
```
Slendy, line 74:
```suggestion
| LBP3     | Mostly compatible      | Compatible                                | Incompatible                  |
```
Also, why is lbp3 for ps3 still mostly compatible instead of fully compatible?

Slendy (changes requested):

README.md
m88youngling: I guess it depends on your definition of compatible vs mostly compatible, as I can see arguments for both definitions. If we define 'compatible' as being playable without crashing, then yes I'd say it's compatible. If we define 'compatible' as meaning feature parity, then I would definitely say it's 'mostly compatible', as LBP3 is still missing a few features. Based on my personal experience, LBP3 is much more stable on Lighthouse than it used to be and I haven't encountered any crashes or issues out of the ordinary. Most of the time it's just LBP3 crapping itself all on its own

Slendy: Yeah if we go by the definition of feature parity then lbp2 shouldn't be marked as compatible either and RPCS3 vs PS3 wouldn't make a difference since it goes off in-game features. I think it should just be compatible for both as well.

TorutheRedFox: compatibility should be defined as all the important features are implemented

LBP3 still has a few important things that aren't implemented yet, at least on main

Ezoiar: ah crap, :skull:

Ezoiar:
I forgot to set the other one as well lol.

Slendy (approved):

PR #524, opened 2022-10-18
Allow RegistrationTokens to have a username pre-selected

Slendy:
This adds a new string to the RegistrationTokens table that requires the user to register with the pre-selected username. In the wise words of Micky Mouse, "It's a surprise tool that will help us later".

Also fixes a bug in BinaryReaderExtensions and makes ChildSlotId not nullable in Score.


daggintosh (approved):
This doesn't appear to cause conflicts and the child slot change is not destructive, 

approved!

Issue #292, opened 2022-04-23
Announce new Team Picks via the Discord webhook


Arcadius2006:
The Discord webhook system should announce when a level is team picked

Slendy:
Closed in #528

PR #528, opened 2022-10-20
Add webhook message when a new Team Pick is added

aMySour:
#292


ProjectLighthouse.Servers.Website/Controllers/Admin/ModerationSlotController.cs
```diff
@@ -27,6 +28,16 @@ public async Task<IActionResult> TeamPick([FromRoute] int id)
         if (slot == null) return this.NotFound();
```
Slendy, line 28: GitHub doesn't let me mark the correct line but on line 27 it should be 
`Slot? slot = await this.database.Slots.Include(s => s.Creator).FirstOrDefaultAsync(s => s.SlotId == id);`
Note the Include() call to fetch the User foreign key, after that you should be able to use slot.Creator.Username in the webhook

ProjectLighthouse.Servers.Website/Controllers/Admin/ModerationSlotController.cs
```diff
@@ -27,6 +28,16 @@ public async Task<IActionResult> TeamPick([FromRoute] int id)
...
+        // Send webhook with slot.Name and slot.Creator.Username
+        if (slot.Creator != null)
+        {
+            await WebhookHelper.SendWebhook("Team Pick", $"Level {slot.Name} by {slot.Creator.Username} has been marked as a team pick.");
+        }
+        else
+        {
+            await WebhookHelper.SendWebhook("Team Pick", $"Level {slot.Name} has been marked as a team pick.");
+        }
+
```
Slendy, line 40:
```suggestion
        await WebhookHelper.SendWebhook("New Team Pick!", $"The level [**{slot.Name}**]({ServerConfiguration.Instance.ExternalUrl}/slot/{slot.SlotId}) by **{slot.Creator?.Username}** has been team picked");
```
Since we used Include() in the slot fetch above there is never a case where the Creator can be null so we don't need the if statement anymore.
Another small nitpick is the message could have a link to the slot that got team picked and I also added emphasis to the creator's name by bolding it

Slendy (changes requested):
Thank you for your contribution to Lighthouse! There are just a few small nitpicks I have regarding how EntityFramework foreign keys get fetched and the phrasing and formatting of the team pick message.

ProjectLighthouse.Servers.Website/Controllers/Admin/ModerationSlotController.cs
aMySour: Fixed

Slendy (approved):
LGTM, thanks for your contribution to Project Lighthouse!

PR #552, opened 2022-11-05
Add Deflate compression for text/xml

Slendy:
Adds support for Deflate compression on text/xml responses that exceed 1kb. This should help with loading times and it is speculated that it can reduce crashing on lbp3 even further (citation needed).


W0lf4llo:
REVIEW: Looks good to me, also please update the branch :D

daggintosh (approved):
Seems fine but citation probably needed on potential benefits.

PR #561, opened 2022-11-15
Fix #545 - test for img size & 1:1 ratio

### LINKED ISSUE #545: Profile picture uploads through the website are not checked for file size or aspect ratio
**Describe the bug**
Profile pictures uploaded through the website are not checked for either file size or aspect ratio, causing potential server storage issues and layout issues. So far I have tested with images of as large as 30MB which is quite unreasonable for a single image.

**To Reproduce**
1. Go to the settings page of any profile you have access to
2. Attempt to upload a profile picture of massive dimensions or a strange aspect ratio
3. Save the profile picture
4. You will get some kind of unexpected behavior; if the page loads after the image is uploaded to the server, you'll see the site cropping the profile picture in some areas and displaying it completely in others. 

**Expected behavior**
The site should have the ability to respond with an error informing the user that their image does not meet certain requirements, like a file size limit or acceptable aspect ratio like 1:1 (or ideally a cropping dialogue would be offered). 

**Screenshots**
image

**Environment Details**
 - OS: Fedora Workstation 36
 - Browser and Version: Firefox (RPM) 106.0.3

ruvaldak:
This is completely untested because I don't have an environment set up to test this with. Just a super quick and dirty fix. Could also be adjusted to, instead of checking dimensions, check actual file size using HttpPostedFileBase.ContentLength

### DIFF
--- a/ProjectLighthouse.Servers.Website/Pages/UserSettingsPage.cshtml
+++ b/ProjectLighthouse.Servers.Website/Pages/UserSettingsPage.cshtml
@@ -143,7 +143,7 @@ function onSubmit(e){
       const file = e.target.files.item(0);
       const img = System.Drawing.Image.FromFile(file);
       const minSize = 1;
-      const maxSize = 1024;
+      const maxSize = 512;
       if (file.type !== "image/jpeg" && file.type !== "image/png")
           return;
       if (!(img.Height <= maxSize && img.Height >= minSize) && (img.Width !== img.Height))
--- a/ProjectLighthouse.Servers.Website/Pages/UserSettingsPage.cshtml
+++ b/ProjectLighthouse.Servers.Website/Pages/UserSettingsPage.cshtml
@@ -141,14 +141,12 @@ function onSubmit(e){
     let selectedAvatar = "";
     document.getElementById("avatar").addEventListener("change", e => {
       const file = e.target.files.item(0);
-      Image img = System.Drawing.Image.FromFile(file);
-      int width = img.Width;
-      int height = img.Height;
-      int minSize = 1;
-      int maxSize = 1024;
+      const img = System.Drawing.Image.FromFile(file);
+      const minSize = 1;
+      const maxSize = 1024;
       if (file.type !== "image/jpeg" && file.type !== "image/png")
           return;
-      if (!(height <= maxSize && height >= minSize) && (width !== height))
+      if (!(img.Height <= maxSize && img.Height >= minSize) && (img.Width !== img.Height))
           return;
       
       const output = document.getElementById('userPicture');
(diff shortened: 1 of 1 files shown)


ProjectLighthouse.Servers.Website/Pages/UserSettingsPage.cshtml
```diff
@@ -140,8 +141,13 @@ function onSubmit(e){
     let selectedAvatar = "";
     document.getElementById("avatar").addEventListener("change", e => {
       const file = e.target.files.item(0);
+      const img = System.Drawing.Image.FromFile(file);
```
Slendy, line 144: This won't work because you're mixing client-side JavaScript with server-side C#. This either needs to be refactored to only use JS code or the image dimension check should be done on the server. Personally, I prefer option 2 because it prevents people from bypassing the client-side check by just crafting their own HTTP request.

ProjectLighthouse.Servers.Website/Pages/UserSettingsPage.cshtml
```diff
@@ -140,8 +141,13 @@ function onSubmit(e){
...
+      const minSize = 1;
+      const maxSize = 512;
       if (file.type !== "image/jpeg" && file.type !== "image/png")
           return;
+      if (!(img.Height <= maxSize && img.Height >= minSize) && (img.Width !== img.Height))
```
Slendy, line 149: This restricts images to a perfect 1:1 aspect ratio and most images aren't perfect squares so a lot of people would have to edit their images to actually use them. This could be changed to allow a 1:1 aspect ratio plus or minus some deviation, or it could be removed entirely to allow the website's CSS to properly scale the image.

Slendy (changes requested):
This needs some more work before it can be merged. I would recommend implementing these checks on the server side rather than on the client as well as fixing the aspect ratio problem by modifying the website's CSS instead of enforcing the uploaded file's dimensions.

ProjectLighthouse.Servers.Website/Pages/UserSettingsPage.cshtml
Arcadius2006: doesn't LBP actually allow non 1:1 images to be used as PFPs? i don't see why the pictures submitted have to be converted to 1:1 if the game already handles everything

Slendy:
Just a heads up, I'm going to be closing this pull request if I don't hear back from you before a month from my review (Jan 1, 2023). If you implement my changes and are ready for review again feel free to open another pull request.

ruvaldak:
Go ahead and close it. Not sure if I'll get around to improving the PR according to your review any time soon.

PR #562, opened 2022-11-18
Add support for Cross Controller Planets

daggintosh:
#### And thus, balance is restored to the LittleBigPlanet community.

This fixes the problem where you can't apply custom planets with LBP2 Cross Controller active, happy creating!

### DIFF
--- a/ProjectLighthouse/PlayerData/Profiles/User.cs
+++ b/ProjectLighthouse/PlayerData/Profiles/User.cs
@@ -137,6 +137,9 @@ public class User
     public string PlanetHashLBPVita { get; set; } = "";
 
     [JsonIgnore]
+    public string PlanetHashCrossControl { get; set; } = "";
+
+    [JsonIgnore]
     public int Hearts => this.database.HeartedProfiles.Count(s => s.HeartedUserId == this.UserId);
 
     [JsonIgnore]
@@ -205,6 +208,7 @@ public class User
                           true
                       ) + // technically not a part of the user but LBP expects it
                       LbpSerializer.StringElement<int>("heartCount", this.Hearts, true) +
+                      this.serializeCrossControlPlanet(gameVersion) +
                       this.serializeEarth(gameVersion) +
                       LbpSerializer.StringElement<string>("yay2", this.YayHash, true) +
                       LbpSerializer.StringElement<string>("boo2", this.BooHash, true) +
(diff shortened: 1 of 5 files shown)


ProjectLighthouse/PlayerData/Profiles/User.cs
```diff
@@ -136,6 +136,9 @@ private int PhotosWithMe()
     [JsonIgnore]
     public string PlanetHashLBPVita { get; set; } = "";
 
+    [JsonIgnore]
+    public string PlanetHashCrossControl { get; set; } = "";
```
Slendy, line 140: This name sounds extra long, maybe it could be reduced to something like `PlanetHashCC`. It kind of removes the implied usage though so it might be better to keep this.

ProjectLighthouse/PlayerData/Profiles/UserUpdate.cs
```diff
@@ -18,6 +18,9 @@ public class UserUpdate
     [XmlElement("planets")]
     public string? PlanetHash { get; set; }
 
+    [XmlElement("crossControlPlanet")]
+    public string? LBP2CrossControlPlanetHash { get; set; }
```
Slendy, line 22: This name is definitely too long, I think it would be better if you dropped the `LBP2` at the beginning or renamed it to `PlanetHashCrossControl` or `PlanetHashCC` like the user

ProjectLighthouse/PlayerData/Profiles/User.cs
```diff
@@ -225,6 +229,12 @@ public string Serialize(GameVersion gameVersion = GameVersion.LittleBigPlanet1)
         return LbpSerializer.TaggedStringElement("user", user, "type", "user");
     }
 
+    private string serializeCrossControlPlanet(GameVersion gameVersion)
+    {
+        if (gameVersion != GameVersion.LittleBigPlanet2) return "";
+        return LbpSerializer.StringElement("crossControlPlanet", this.PlanetHashCrossControl);
```
Slendy, line 235: this could probably go inside serializeEarth, if you think of a better implementation than what I have here feel free to use it
```c#
    private string serializeEarth(GameVersion gameVersion)
    {
        string planet = LbpSerializer.StringElement<string>
        (
            "planets",
            gameVersion switch
            {
                GameVersion.LittleBigPlanet2 => this.PlanetHashLBP2,
                GameVersion.LittleBigPlanet3 => this.PlanetHashLBP3,
                GameVersion.LittleBigPlanetVita => this.PlanetHashLBPVita,
                _ => "", // other versions do not have custom planets
            },
            true
        );
        if(gameVersion == GameVersion.LittleBigPlanet2) 
            planet += LbpSerializer.StringElement<string>("crossControlPlanet", this.PlanetHashCrossControl, true);
        return planet;
    }
```

Slendy (changes requested):
Looks good, just some small things about naming and putting the serialization into the planet serializer function

Slendy (approved):
LGTM

PR #567, opened 2022-11-28
Added three new categories to LBP3

W0lf4llo:
Adds Most Hearted, Most Played, and Highest Rated category to LBP3


Slendy:
Aren't these handled by the organize button?

W0lf4llo:
> Aren't these handled by the organize button?
> 

Yes but the organize button isn’t functional yet. And these tabs are still included in the official servers. It’s also much more convenient and easier to access then going to newest levels, organize tab, and filtering them out. Also doesn’t confuse new players :)

Slendy (approved):
Approving for now but this might be removed when the community tab is reimplemented.

PR #595, opened 2022-12-16
create more censor filter configuration options

sudokoko:
- **added screaming:** 3 screaming strings, `FilterMode.Scream`
- **added half-life quotes:** 13 half-life quote strings, `FilterMode.HalfLife`
- bumped config version to 15 as well


Slendy (changes requested):
Thanks for your contribution to Project Lighthouse. After review, your code looks good but this feature is unnecessary in my opinion. This PR would be better suited if it allowed instance owners to have a custom filter mode that let them specify replacement words in the config. If you need inspiration for things that should be implemented or need to be fixed check the issues page.

sudokoko:
i'm going to go ahead and convert this to a draft to potentially come back to later. thanks though :)

PR #599, opened 2022-12-17
Separate user ban check into its own statement

sudokoko:
* Separating the `user.IsBanned` check from the `user == null` check for clarity
  * Was previously combined (`user == null || user.IsBanned`)
  * While this does work, having the same error for both checks can present issues.
    * Both checks display the error `Unable to find user {npTicket.Username} from token`.
    * This can cause issues when trying to troubleshoot problems.


m88youngling:
Closed for redundancy

PR #600, opened 2022-12-17
Rework login and registration systems

### LINKED ISSUE #569: NPTicket signature verification
NPTickets, which are used by the game to authenticate with the game server, contain a cryptographic signature of its contents. This signature can be used to verify that a ticket was indeed generated by the expected source and hasn't been tampered with. Fortunately for us, RPCN is open source which means verifying those tickets is relatively simple. The challenge is verifying tickets generated by PSN because there is no official documentation and the tickets are generated server-side so you can't simply reverse engineer the console firmware.

# What is the signature?
`30 36 02 19 00 CC 99 5F 2D 33 4A 4C 2B AD BB 07 1D 21 DF A1 05 BF 12 DC F1 58 07 E8 41 02 19 00 D3 A3 AA C3 4A 20 23 D9 04 2A 7B F2 E3 5A DE 2F E5 B5 41 BA 48 13 37 41`
The signature is located in the footer section of the ticket and is two DER-encoded integers that are roughly 192 bits in length. I believe that this is an ECDSA signature and the two integers correspond to r and s respectively. Theoretically, all we would need to verify this signature is the public key of the signer which in this case is PSN. Due to the nature of ECC, it is possible to recover the public key from just the signature and message but this requires knowledge of the elliptic curve properties, the hashing algorithm, and the exact message bytes.

# What hasn't worked
I have already tried using common 192 bits curves like `secp192k1` and `secp192r1` with common hash algorithms like `SHA1`, `SHA224`, and `SHA256`. Using a sample size of a few hundred tickets I was unable to produce any results using [a small C# program I wrote](https://github.com/Slendy/PubKeyFinder). As a proof of concept, this program is able to recover the public key of signatures from RPCN which means that PSN tickets are either using an uncommon or nonstandard curve type, are using an uncommon hashing algorithm, or the hashed message is more than just the body section of the ticket verbatim. 

# Why does this matter?
If we could cryptographically verify NPTickets then it would mean that Project Lighthouse would no longer have to rely on an IP-based authentication scheme which can be a source of frustration for users with dynamic IP addresses. It could also light the way for a system that is more akin to the original servers where you don't even have to register to connect to the game server. Looking beyond Project Lighthouse, advancements in ticket verification could benefit other private server projects and make them more secure.

# What can I do?
If you read this far and aren't a programmer or cryptography guru and you know anyone who is familiar with the inner workings of PSN, NPTickets or is knowledgeable of elliptic curve cryptography and ECDSA, then link them to this issue and ask if they can help. This is a somewhat complex issue and we could use all the help we can get.

Slendy:
Removes the reliance on IP addresses to enforce account security for the game server. This is achieved by cryptographically verifying the login tickets to ensure that they originate from either PSN or the official RPCN server.

## New registration flow
Users will no longer have to register on the website to play on the servers. Now if their name is available an account will automatically be created for them when they try and connect for the first time. Obviously, this leaves gaps like how do you register for the website if it creates an account for you? That is handled by users setting their email in-game and then requesting a password reset on the website. This could be reworked to automatically send an email when they set it as well.

## Account linking
For users that already have an account, this update will now require them to link their platforms (either PSN or RPCN). This means that if a username is owned by a separate PSN and RPCN user only one of them will be able to have an account. It is possible to link an account to both platforms if you switch between PS3 and RPCS3 and have the same name. To link an account you simply try to log in as before and it will still give you a 403 error but once you approve the link request you will never have to approve a request again even if your IP address changes. 
image
image

Closes #569


daggintosh (approved):
excellent, thank you for for implementing the **Rework login and registration systems #600**

Issue #535, opened 2022-10-25
Custom Word Filter


m88youngling:
@Slendy mentioned to me that a word filter can be enabled with Lighthouse but all the words are hardcoded. It'd be very handy to have a customizable word filter. You could add words to a comma separated list either in the lighthouse.yml config or you could customize the word filter via the admin panel.

- [ ] Filter level titles
- [ ] Filter comments
- [ ] Filter reviews
- [ ] Filter user profiles

Zaprit:
currently it's [here](https://github.com/LBPUnion/ProjectLighthouse/blob/main/ProjectLighthouse/chatCensoredList.txt)

m88youngling:
> currently it's [here](https://github.com/LBPUnion/ProjectLighthouse/blob/main/ProjectLighthouse/chatCensoredList.txt)

Can instance owners easily edit the list on their server, or will it auto update when they update to latest changes from the repo?

Zaprit:
it will likely be clobbered when the server gets updated

m88youngling:
Crap 💀 how do we deal with that?

Zaprit:
give me an hour

m88youngling:
I have the 'furry' mode activated on Beacon and it doesn't seem to moderate level titles

m88youngling:
Seems that comments don't get filtered either, I bet that only in-game text chat and magic mouths get filtered at the moment

Ezoiar:
To avoid language collisions, I suggest using the selected/preferred language as well to select a specific group of word filter.
For example, using fantasy made up languages lol: Frunkish: Ung ba moda (Fu*k you man) Mongtsh: Ung ta te moda (How is your day?). Right now looking at the report at #bug-blasters from October 29th, I assume right now it goes like this.

Ung ba moda - gets censored because it means something bad
Ung ta te moda - gets censored anyways despite not being bad.

With the solution I suggested, It goes like this:

User with Frunkish set as preferred language:

Ung ba moda (gets censored)
Ung ta te moda (gets censored)

User with Mongtsh set as preferred language:

Ung ba moda (does not get censored)
Ung ta te moda (does not get censored)

This solution to a problem does bring both good and potential bad.
The goods: Brings more freedom to foreign-speaking users and community, Does not censor any word in foreign languages.
The bad, there's only one: If the user does abuse the solution by speaking English despite being on foreign language, Only the host/other players can report it via good grief, This could be a violation for circumventing the word filter.

[...]

PR #603, opened 2022-12-20
Create .gitattributes

Cryptofowl:
Added a .gitattributes file that excludes the local customWordFilter.txt file from merged updates, allowing server operators to maintain their own word filter list and not have it overwritten. For issue #535


m88youngling:
It may be useful to, in the future, have separate 'announcements' and 'eula' files that lighthouse.yml can pull from instead of directly typing the stuff into the config. This PR would be useful for that as well later on

Slendy (changes requested):
First of all, thank you for your contribution to Project Lighthouse! This is definitely a good way to solve #535. Unfortunately, this does have the side effect of causing the git branch to be marked as dirty because it tracks chatCensoredList.txt as an uncommitted file. I think you could solve this by maybe adding it to the .gitignore?

Cryptofowl:
Good point! I will give that a try

Slendy:
For some reason, even with the file added to the gitignore it is still being tracked and has to be manually ignored with `git update-index --assume-unchanged`. I also tested merging; it doesn't automatically merge with your local changes or even discard the remote changes. It forces you to discard your local changes or commit them. It may just be easier to move the censor list to a dynamic file that's loaded at runtime similar to the config.

Cryptofowl:
Well damn. Alright, that makes sense. So, as I understand it, that would involve changes to the censor helper similar to how lighthouse.yml is generated by serverConfiguration.cs?

Cryptofowl:
Changes made, review (strongly) recommended because I've never used C# or asp and had a few uncertainties. Mostly just followed the way this was implemented for the server config file.

Cryptofowl:
I also went ahead and added filtering for all the remaining categories in #535. Assuming that I'm understanding things correctly and not breaking things, this should about do it.

.gitignore
```diff
@@ -36,6 +36,7 @@ logs/*
 npTicket*
 r.tar.gz
 lighthouse.yml.configme
+chatCensoredList.txt
```
Slendy, line 39:
```suggestion
```
If we aren't doing the git method anymore the chatCensoredList doesn't need to be in the gitignore and you could also remove the file from the project altogether since it would be handled by the code.

ProjectLighthouse.Servers.GameServer/Controllers/CommentController.cs
```diff
@@ -119,7 +119,9 @@ public async Task<IActionResult> PostComment(string? username, string? slotType,
             targetId = await this.database.UserIdFromUsername(username!);
         }
 
-        bool success = await this.database.PostComment(token.UserId, targetId, type, comment.Message);
+        string scannedText = CensorHelper.ScanMessage(comment.Message);
```
Slendy, line 122:
```suggestion
        string filteredText = CensorHelper.ScanMessage(comment.Message);
```
I think the scanned verb sounds a little weird since we're actually changing the message here

ProjectLighthouse/Helpers/CensorHelper.cs
```diff
@@ -18,6 +20,67 @@ public static class CensorHelper
...
+
+    private void writeConfig(string path)
+    {
+        ISerializer serializer = new SerializerBuilder().WithNamingConvention(namingConvention).Build();
+
+        File.WriteAllText(path, serializer.Serialize(defaultCensoredList));
+    }
+
+    }
+
```
Slendy, line 83:
```suggestion
```
What you're trying to do here is basically copy the already existing configuration system, so it would make more sense to move this to a `CensorConfiguration` class in `LBPUnion.ProjectLighthouse.Configuration.ConfigurationCategories` then the filter mode and the word list could be in the same section of the lighthouse.yml config like this:
image
<details>
<summary>Code</summary
</details>

```c#
public class CensorConfiguration
{
    public FilterMode UserInputFilterMode { get; set; } = FilterMode.None;

    public List<string> CensorList { get; set; } = new()
    {
        "cunt",
        "fag",
        "faggot",
        "horny",
        "kook",
        "kys",
        "loli",
        "nigga",
        "nigger",
        "penis",
        "pussy",
        "retard",
        "retarded",
        "vagina",
        "vore",
        "restitched",
        "h4h",
    };
}
```

Slendy (changes requested):
The comment, review, and publishing filters are fine. What you are trying to do with CensorHelper could probably be better achieved by adding new options to the lighthouse.yml config. You could also probably remove the old chatCensoredList.txt file since it would be hardcoded into the configuration anyways.

ProjectLighthouse/Helpers/CensorHelper.cs
Zaprit: surely putting the word list in the lighthouse.yml is going to make it a nightmare to actually look through the config file. Maybe as the scope of our config expands we should consider making multiple config files, such as

- a listen configuration for the 3 servers
- database/redis
- monitoring
- word filters
- authentication

this way the config files become more readable and you don't have to scroll through a 400 line word filter to change the instance name (though this may be better suited for another issue)

ProjectLighthouse.Servers.GameServer/Controllers/CommentController.cs
Cryptofowl: I agree, I was mainly trying to keep it consistent with the existing naming convention in the message filtering controller. Change both maybe?

ProjectLighthouse/Helpers/CensorHelper.cs
Slendy: If we do split the configuration file then the current system needs to be refactored to more easily allow for that rather than duplicating the same class multiple times

Cryptofowl: Alright so it sounds like I should go ahead remove this code from this commit so we can get this PR merged. Then we can open up another issue for refactoring the config?

Slendy: I have the config abstraction working in #624 but it's waiting for review

Slendy:
Just letting you know that the new configuration system has been merged. Feel free to reach out if you run into any problems.

Cryptofowl:
> The comment, review, and publishing filters are fine. What you are trying to do with CensorHelper could probably be better achieved by adding new options to the lighthouse.yml config. You could also probably remove the old chatCensoredList.txt file since it would be hardcoded into the configuration anyways.

With the new config system merged, looks like this is good to go

Slendy:
Were you still planning to move the word list into a separate config file?

Cryptofowl:
Ah, right, sorry I thought that was covered in your refactor. I will get that added

Slendy (approved):
LGTM, I went ahead and made the necessary changes for having a separate configuration file

PR #614, opened 2022-12-29
remove redundant case switches

sudokoko:


sudokoko:
trying a different method, hold

Slendy:
Closing as this is 1. Not meaningful, 2. Not necessarily correct and 3. Rider complains if you don't handle all possible enum values in a switch statement. Having all possible cases also has the benefit of being more verbose and self-documenting.

PR #622, opened 2023-01-05
Remove duplicate scores

### LINKED ISSUE #607: Redundancy In Displaying Scores on Website
When a group of players submit a score, it resubmits the score for x amount of times (x = the amount of players) 

image

The most likely way of fixing this issue is to only accept the score of the host. Determining who's the host should be relatively simple, as it should use the same method used for Good Grief reports.

Slendy:
This reverts the change that submits scores for all players instead of just from the host. It also runs a migration to remove all duplicate scores. Still needs to be tested on a clone of production to ensure accuracy. Closes #607


sudokoko (approved):
LGTM

daggintosh (approved):
lgtm

PR #624, opened 2023-01-07
Abstract config design and update logging format

Slendy:
This PR makes the config system way more abstract which should make it easier to add new config files in the future. It also drops legacy config support. I assume most people aren't going to be running the old JSON style config this late into development but at worst they will have to manually transfer their settings.


daggintosh (approved):
looks ok

PR #647, opened 2023-01-27
Fix comments not being disabled if banned

sudokoko:
Hotfix for comments not actually being disabled if a user is banned (whoops)


daggintosh (approved):
Hotfix

Zaprit (approved):
LGTM

PR #648, opened 2023-01-27
Fix bug where users can't be deleted

### LINKED ISSUE #643: Attempting to delete a user results in an error code 500.
**Describe the bug**
Whenever I attempt to delete a user from the website via the admin panel, the resulting page is an error code 500.

**To Reproduce**
1. Go to the admin panel. 
2. Type in the user you want to delete. 
3. Press enter or click submit. 
4. The resulting page is an error code 500 and the user will not be deleted. 

**Expected behavior**
I believe the resulting page is a refreshed admin panel with a success console message at the top of the screen. The user is then deleted. 

**Additional context**
It's sometimes necessary for us to delete accounts in the event of user request or if a test account needs to be deleted.

Slendy:
This bug was due to the fact that the case system is designed around requiring the case creator/case dismisser to exist. When you would try to delete a user that occupied either of these fields it would result in an error. This fix is definitely not the optimal solution since it uses the placeholder user but I don't have time to rewrite the entire case system. The new functionality is that when you delete a user who dismissed/created a case it will still show their username but it won't be a clickable link.
Closes #643

### DIFF
--- a/.config/dotnet-tools.json
+++ b/.config/dotnet-tools.json
@@ -3,7 +3,7 @@
   "isRoot": true,
   "tools": {
     "dotnet-ef": {
-      "version": "7.0.1",
+      "version": "7.0.2",
       "commands": [
         "dotnet-ef"
       ]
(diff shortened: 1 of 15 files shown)


daggintosh (approved):
Seems fine to me. Maybe a third set of eyes would spot any remaining issues if I missed any.

PR #650, opened 2023-01-27
useless, as those get caught by default

McMistrzYT:
removed some cases in the user controller as they change absolutely nothing and get caught by default anyway


sudokoko:
Wasn't this occurrence previously discussed? Feel free to correct me if I'm wrong but I believe Slendy explained why they're there.

McMistrzYT:
they do absolutely nothing because default is right under them no?

sudokoko:
Let me quickly find his explanation, moment.

sudokoko:
This was similar to or the same reason #614 was closed.

> Rider complains if you don't handle all possible enum values in a switch statement. Having all possible cases also has the benefit of being more verbose and self-documenting.

McMistrzYT:
makes sense i guess

PR #652, opened 2023-01-29
Major refactor and reorganization of types

Slendy:
This PR reorganizes most of the type classes and enums within Lighthouse. It organizes all of the DbSet entity types into a single folder and other types are organized into ProjectLighthouse/Types. Some types are only used by a single server and as such, they exist in the corresponding server's Types folder.


ProjectLighthouse/Database.cs
```diff
@@ -3,56 +3,56 @@
...
+using LBPUnion.ProjectLighthouse.Types.Misc;
 using Microsoft.AspNetCore.Http;
 using Microsoft.EntityFrameworkCore;
 
 namespace LBPUnion.ProjectLighthouse;
 
 public class Database : DbContext
 {
+    // ReSharper disable once InconsistentNaming
+    public DbSet<ApiKey> APIKeys { get; set; }
```
Zaprit, line 25: the variable should also be called ApiKeys, no?

Slendy: I tried to avoid doing a migration in this PR so it has to stay capitalized to match the actual table name

Zaprit: fair enough

Zaprit (commented):
Pretty cool, that was a lot of files

Slendy:
LGTM
image

PR #656, opened 2023-02-05
Make rate limiter match zone rather than directly comparing url

Slendy:
Fixes the built-in rate limiter of Lighthouse to use regex matching to count the number of requests the user has made for a specific zone. The old implementation would just count the number of requests that matched the URL exactly which meant for a rate limit zone over /user/* someone could exceed the rate limit at /user/15 and then exceed the rate limit again at /user/16. This is now prevented because all requests are properly counted.

Also adds a page for when you reach the rate limit rather than just letting the browser give a default 429 error.
image


daggintosh (approved):
Switch expressions look very interesting.

This seems fine.

sudokoko (approved):
LGTM

PR #662, opened 2023-02-08
Profile Blocking

Zaprit:
Allows users to block people they don't like.

Adds a block button to the users page


ProjectLighthouse/Database.cs
```diff
@@ -10,6 +10,7 @@
 using LBPUnion.ProjectLighthouse.Helpers;
 using LBPUnion.ProjectLighthouse.Levels;
 using LBPUnion.ProjectLighthouse.Levels.Categories;
+using LBPUnion.ProjectLighthouse.Logging;
```
Slendy, line 13:
```suggestion
```
I think this might be a leftover from debug logging

ProjectLighthouse/Database.cs
```diff
@@ -53,7 +54,7 @@ public class Database : DbContext
     public DbSet<APIKey> APIKeys { get; set; }
     public DbSet<Playlist> Playlists { get; set; }
     public DbSet<PlatformLinkAttempt> PlatformLinkAttempts { get; set; }
-
+    public DbSet<BlockedProfile> BlockedProfiles { get; set; }
```
Slendy, line 57:
```suggestion
    public DbSet<BlockedProfile> BlockedProfiles { get; set; }
    
```
I would leave an extra line between the dbsets and the next function

ProjectLighthouse/Database.cs
```diff
@@ -131,9 +132,11 @@ public async Task<bool> RateComment(int userId, int commentId, int rating)
         Comment? comment = await this.Comments.FirstOrDefaultAsync(c => commentId == c.CommentId);
 
         if (comment == null) return false;
-
+
```
Slendy, line 135:
```suggestion

```
just trim the whitespace here

ProjectLighthouse/Database.cs
```diff
@@ -312,6 +317,40 @@ public async Task UnqueueLevel(int userId, Slot queuedSlot)
...
+        this.BlockedProfiles.RemoveWhere(bp => bp.BlockedUser == blockedUser && bp.UserId == userId);
+    }
+
+    public async Task<bool> IsBlocked(int blockedUserId, int userId)
+    {
+        if (userId == blockedUserId) return false;
+
+        return await this.BlockedProfiles.Has(bp => bp.BlockedUserId == blockedUserId && bp.UserId == userId);
+    }
```
Slendy, line 352:
```suggestion
    public async Task<bool> IsUserBlockedBy(int userId, int targetId)
    {
        if (userId == targetUser) return false;

        return await this.BlockedProfiles.Has(bp => bp.BlockedUserId == userId && bp.UserId == targetUser);
    }
```
The IsBlocked naming is kind of ambiguous as to what it is checking in relation to the parameters. I think a more descriptive name could help a little along with renaming the parameters to go along with the name.

ProjectLighthouse/Database.cs
```diff
@@ -312,6 +317,40 @@ public async Task UnqueueLevel(int userId, Slot queuedSlot)
...
+        await this.BlockedProfiles.AddAsync(blockedProfile);
+
+        await this.SaveChangesAsync();
+    }
+
+    public void UnblockUser(int userId, User blockedUser)
+    {
+        if (userId == blockedUser.UserId) return;
+
+        this.BlockedProfiles.RemoveWhere(bp => bp.BlockedUser == blockedUser && bp.UserId == userId);
```
Slendy, line 344: I'd have to look at the generated SQL, but it might be more efficient to compare against the user's id rather than the foreign key field.

ProjectLighthouse/Database.cs
```diff
@@ -312,6 +317,40 @@ public async Task UnqueueLevel(int userId, Slot queuedSlot)
         await this.SaveChangesAsync();
     }
 
+    public async Task BlockUser(int userId, User blockedUser)
+    {
+        if (userId == blockedUser.UserId)
+        {
+            return;
+        }
```
Slendy, line 325:
```suggestion
        if (userId == blockedUser.UserId) return;
```
I would inline this to match the styling in `UnblockUser`

ProjectLighthouse/Database.cs
```diff
@@ -312,6 +317,40 @@ public async Task UnqueueLevel(int userId, Slot queuedSlot)
...
+        
+        User? user = await this.Users.FirstOrDefaultAsync(u => u.UserId == userId);
```
Slendy, line 327:
```suggestion
        User? user = await this.Users.FindAsync(u => u.UserId == userId);
```
I've been meaning to change a lot of the FirstOrDefault calls that don't use .Include because FindAsync can apparently work better for finding entities by primary key and can reduce database trips if the entity is already tracked.

ProjectLighthouse/Database.cs
```diff
@@ -312,6 +317,40 @@ public async Task UnqueueLevel(int userId, Slot queuedSlot)
...
+        
+        BlockedProfile blockedProfile = new()
+        {
+            User = user,
+            BlockedUser = blockedUser,
```
Slendy, line 332: Does this properly set the UserId and BlockedUserId fields? Generally, I like to set the actual fields that will update in the db but if it works then I guess it's fine.

ProjectLighthouse/Database.cs
```diff
@@ -189,14 +192,16 @@ public async Task<bool> PostComment(int userId, int targetId, CommentType type,
...
         {
-            int targetSlotId = await this.Slots.Where(s => s.SlotId == targetId)
+            Slot? targetSlot = await this.Slots.Where(s => s.SlotId == targetId)
                 .Where(s => s.CommentsEnabled && !s.Hidden)
-                .Select(s => s.SlotId)
                 .FirstOrDefaultAsync();
-            if (targetSlotId == 0) return false;
+            if (targetSlot == null) return false;
+            
+            if (await this.IsBlocked(userId, targetSlot.CreatorId)) return false;
```
Slendy, line 204:
```suggestion
            if (await this.IsUserBlockedBy(userId, targetSlot.CreatorId)) return false;
```
To go with the function rename

ProjectLighthouse/Database.cs
```diff
@@ -189,14 +192,16 @@ public async Task<bool> PostComment(int userId, int targetId, CommentType type,
...
         }
         else
         {
-            int targetSlotId = await this.Slots.Where(s => s.SlotId == targetId)
+            Slot? targetSlot = await this.Slots.Where(s => s.SlotId == targetId)
                 .Where(s => s.CommentsEnabled && !s.Hidden)
-                .Select(s => s.SlotId)
                 .FirstOrDefaultAsync();
-            if (targetSlotId == 0) return false;
+            if (targetSlot == null) return false;
```
Slendy, line 202: You can be cheeky here and select the slot CreatorId instead of the SlotId because if the slot doesn't exist it will return 0 regardless of the Select but by projecting the creator instead of the SlotId you save having to load the entire object.

ProjectLighthouse.Servers.Website/Pages/UserPage.cshtml
```diff
@@ -61,6 +61,20 @@
         <br>
         @if (Model.ProfileUser != Model.User && Model.User != null)
         {
+            if (!Model.IsProfileUserBlocked)
+            {
+                <a class="ui red button" href="/user/@Model.ProfileUser.UserId/block">
+                    <i class="ban icon"></i>
```
sudokoko, line 67: Just a UI suggestion to match the Volume Up icon you have below.
```suggestion
                    <i class="volume mute icon"></i>
```

ProjectLighthouse.Servers.Website/Controllers/UserPageController.cs
```diff
@@ -87,4 +87,32 @@ public async Task<IActionResult> UnheartUser([FromRoute] int id)
...
+        WebToken? token = this.database.WebTokenFromRequest(this.Request);
+        if (token == null) return this.Redirect("~/login");
+
+        User? blockedUser = await this.database.Users.FirstOrDefaultAsync(u => u.UserId == id);
+        if (blockedUser == null) return this.NotFound();
+
+        this.database.UnblockUser(token.UserId, blockedUser);
+
+        return this.Redirect("~/user/" + id);
+    }
```
Slendy, line 117: You could possibly combine this into one function since they share a lot of code but it might be at the expense of an extra db trip. Might be out of the scope of this PR though.

sudokoko (approved):
LGTM
Common Zaprit W :trollface:

ProjectLighthouse.Servers.Website/Pages/UserPage.cshtml.cs
```diff
@@ -88,11 +90,15 @@ public async Task<IActionResult> OnGet([FromRoute] int userId)
...
-            this.Comments = await this.Database.Comments.Include(p => p.Poster)
-                .OrderByDescending(p => p.Timestamp)
-                .Where(p => p.TargetId == userId && p.Type == CommentType.Profile)
-                .Take(50)
-                .ToListAsync();
+            this.Comments = await
+                (from comment in this.Database.Comments
+                where !(
+                    from blockedProfile in this.Database.BlockedProfiles 
+                    where blockedProfile.UserId == this.User.UserId
```
Slendy, line 97: `this.User` can be null (does this work if you aren't logged in) so you might be able to do `this.User?.UserId`. This also looks kind of wacky but if it works and doesn't pull in the entire comment/blocked user table I'm fine with it

Slendy (changes requested):
Few things to change here, mostly related to small semantics/naming. One thing I see missing though is the game server implementation of not including comments from blocked users. Also, this could maybe use a page where you can see all the users you've blocked with quick actions to unblock them.

ProjectLighthouse/Database.cs
Zaprit: Resolved

ProjectLighthouse/Database.cs
Zaprit: This looks wrong, shouldn't it be targetId on line 348

ProjectLighthouse/Database.cs
Zaprit: Who knows, I'd have to look into that

ProjectLighthouse/Database.cs
Zaprit: It worked in my testing

ProjectLighthouse/Database.cs
Slendy: Yeah targetUser should be targetId

ProjectLighthouse/Database.cs
Zaprit: seems good

ProjectLighthouse.Servers.Website/Controllers/UserPageController.cs
Zaprit: I think considering a lot of our bottlenecks seem to be in the DB I'll save the DB trip

ProjectLighthouse/Database.cs
Zaprit: I've done that just in case

Slendy (approved):
LGTM

PR #664, opened 2023-02-09
Remove InfluxDB and add /api/v1/playerCount

### LINKED ISSUE #658: InfluxDB monitoring shouldn’t be handled by Lighthouse
Right now the monitoring system with InfluxDB directly sends data to a bucket rather than using something like Telegraf to ingest data from HTTP endpoints and/or log files. Allowing the server owner to make their own configuration setups seems better than having an option for monitoring which forces specific data points to be measured. There is also conflicts where each server type will try and send Influx monitoring data which can lead to duplicate data points.

Slendy:
This PR removes the hardcoded InfluxDB logging. The functionality can be replicated using the new player count API and telegraf.

An example json response for `/api/v1/playerCount/` is
```json
{"totalPlayerCount":0,"games":[{"game":"LittleBigPlanet1","playerCount":0},{"game":"LittleBigPlanet2","playerCount":0},{"game":"LittleBigPlanet3","playerCount":0},{"game":"LittleBigPlanetVita","playerCount":0},{"game":"LittleBigPlanetPSP","playerCount":0}]}
```

If anyone is curious here is an example telegraf config for fetching and parsing the player data from a lighthouse instance.
```
 [[inputs.http]]
    urls = ["http://localhost:10062/api/v1/playerCount"]
    data_format = "json_v2"
    [[inputs.http.json_v2]]
        measurement_name = "lighthouse"
        [[inputs.http.json_v2.object]]
            path = "games"
            optional = false
            [[inputs.http.json_v2.object.tag]]
                path = "#.game"
                optional = false
            [[inputs.http.json_v2.object.field]]
                path = "#.playerCount"
                rename = "playerCountGame"
                type = "int"
                optional = false
```

Closes #658


sudokoko (approved):
`/api/v1/playerCount` sounds useful, will have to experiment with it.

LGTM.

PR #668, opened 2023-02-11
Feature/user blocking hotfix 1

Zaprit:
This fixes the two issues that blocking was having, mainly comments and reviews on levels still showed on the website and comments in game wouldn't load at all


m88youngling:
Uhhh... LGTM! 😅 (please revert this if it breaks. I'm trying to see if this fix corrects downgraded performance on Beacon with higher user load.)

PR #670, opened 2023-02-12
Highlight moderator/administrator user account links

sudokoko:
This change adds global highlighting of moderator and administrator username links. Moderator users are highlighted orange [`rgb(200, 130, 0)`] and administrator users are highlighted red [`rgb(255, 0, 0)`]. Any links that utilize `ProjectLighthouse.Servers.Website/Pages/Partials/Links/UserLinkPartial.cshtml` will have this highlighting.


Slendy (approved):
LGTM

PR #673, opened 2023-02-12
Fix missing role highlighting on some pages

sudokoko:
Some pages (such as Users Search or the Profile Page) had missing role color highlighting, so I updated the UserCardPartial and ReviewPartial to address this issue.


sudokoko:
Closing, will redo differently later.

PR #679, opened 2023-02-15
Replace Location table with packed 64-bit int

Slendy:
Removes the Location table which provides many benefits:
* Don't have to join another table theoretically improving performance
* Don't have to worry about Location entries potentially going unused, which wastes space.
* Locations have a fallback value now instead of being null/empty.
* Saves space in DB

The new location format is a 64-bit unsigned integer that stores 2 32-bit unsigned integers. Endianness is a concern when we're doing bit operations on numbers like this but 90% of systems nowadays are little-endian and an even smaller subset can actually run .NET 7 so this should be fine in practice.


Slendy:
LGTM. Tested the migration on a staging server, and seems to work fine.

PR #680, opened 2023-02-19
User Search API

Zaprit:
This adds an API endpoint for searching for users based on a query

I have two things in mind for this
 - A more responsive user search in the future for the website
 - User searching/verification in a newer version of CrashHelper

### DIFF
--- a/ProjectLighthouse.Servers.API/Controllers/UserEndpoints.cs
+++ b/ProjectLighthouse.Servers.API/Controllers/UserEndpoints.cs
@@ -57,7 +57,7 @@ public class UserEndpoints : ApiEndpointController
     /// <returns>A list of users</returns>
     /// <response code="200">The list of users, if any were found</response>
     /// <response code="404">No users matched the query</response>
-    [HttpGet("user_search")]
+    [HttpGet("search/user")]
     [ProducesResponseType(typeof(User), StatusCodes.Status200OK)]
     [ProducesResponseType(StatusCodes.Status404NotFound)]
     public async Task<IActionResult> SearchUsers(string query)
--- a/ProjectLighthouse.Servers.API/Controllers/UserEndpoints.cs
+++ b/ProjectLighthouse.Servers.API/Controllers/UserEndpoints.cs
@@ -65,6 +65,7 @@ public class UserEndpoints : ApiEndpointController
         List<User> users = await this.database.Users
             .Where(u => u.PermissionLevel != PermissionLevel.Banned && u.Username.Contains(query))
             .Where(u => u.ProfileVisibility == PrivacyType.All) // TODO: change check for when user is logged in
+            .Take(20)
             .OrderByDescending(b => b.UserId)
             .ToListAsync();
         if (users.Count == 0)
--- a/ProjectLighthouse.Servers.API/Controllers/UserEndpoints.cs
+++ b/ProjectLighthouse.Servers.API/Controllers/UserEndpoints.cs
@@ -51,6 +51,31 @@ public class UserEndpoints : ApiEndpointController
     }
 
     /// <summary>
+    /// Searches for the user based on the query
+    /// </summary>
+    /// <param name="query">The search query</param>
+    /// <returns>A list of users</returns>
+    /// <response code="200">The list of users, if any were found</response>
+    /// <response code="404">No users matched the query</response>
+    [HttpGet("user_search")]
+    [ProducesResponseType(typeof(User), StatusCodes.Status200OK)]
+    [ProducesResponseType(StatusCodes.Status404NotFound)]
+    public async Task<IActionResult> SearchUsers(string query)
+    {
+        List<User> users = await this.database.Users
+            .Where(u => u.PermissionLevel != PermissionLevel.Banned && u.Username.Contains(query))
+            .Where(u => u.ProfileVisibility == PrivacyType.All) // TODO: change check for when user is logged in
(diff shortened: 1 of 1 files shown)


ProjectLighthouse.Servers.API/Controllers/UserEndpoints.cs
```diff
@@ -50,6 +50,32 @@ public async Task<IActionResult> GetUser(string username)
...
     }
 
+    /// <summary>
+    /// Searches for the user based on the query
+    /// </summary>
+    /// <param name="query">The search query</param>
+    /// <returns>A list of users</returns>
+    /// <response code="200">The list of users, if any were found</response>
+    /// <response code="404">No users matched the query</response>
+    [HttpGet("user_search")]
```
Slendy, line 60: Shouldn't this be something like `/user/search`?

ProjectLighthouse.Servers.API/Controllers/UserEndpoints.cs
```diff
@@ -50,6 +50,32 @@ public async Task<IActionResult> GetUser(string username)
...
+    [ProducesResponseType(typeof(User), StatusCodes.Status200OK)]
```
Slendy, line 61:
```suggestion
    [ProducesResponseType(typeof(User[]), StatusCodes.Status200OK)]
```
I don't know if this is required since this is mainly for the swagger self-documenting thing that doesn't even seem to really work... but probably still a good idea to look into because you're returning a list of users

ProjectLighthouse.Servers.API/Controllers/UserEndpoints.cs
```diff
@@ -50,6 +50,32 @@ public async Task<IActionResult> GetUser(string username)
...
+    [ProducesResponseType(StatusCodes.Status404NotFound)]
+    public async Task<IActionResult> SearchUsers(string query)
+    {
+        List<User> users = await this.database.Users
+            .Where(u => u.PermissionLevel != PermissionLevel.Banned && u.Username.Contains(query))
+            .Where(u => u.ProfileVisibility == PrivacyType.All) // TODO: change check for when user is logged in
+            .Take(20)
+            .OrderByDescending(b => b.UserId)
```
Slendy, line 69:
```suggestion
            .OrderByDescending(b => b.UserId)
            .Take(20)
```
The limit command should usually go after the order by but I'm not an SQL expert so take this advice with a grain of salt. EntityFramework might already make this optimization regardless of how you structure the LINQ

ProjectLighthouse.Servers.API/Controllers/UserEndpoints.cs
```diff
@@ -50,6 +50,32 @@ public async Task<IActionResult> GetUser(string username)
...
+            .ToListAsync();
+        if (users.Count == 0)
+        {
+            return this.NotFound();
+        }
```
Slendy, line 74:
```suggestion
        if (!users.Any()) return this.NotFound();
```
This kinda looks better as one line.

ProjectLighthouse.Servers.API/Controllers/UserEndpoints.cs
```diff
@@ -50,6 +50,32 @@ public async Task<IActionResult> GetUser(string username)
...
+    /// <returns>A list of users</returns>
+    /// <response code="200">The list of users, if any were found</response>
+    /// <response code="404">No users matched the query</response>
+    [HttpGet("user_search")]
+    [ProducesResponseType(typeof(User), StatusCodes.Status200OK)]
+    [ProducesResponseType(StatusCodes.Status404NotFound)]
+    public async Task<IActionResult> SearchUsers(string query)
+    {
+        List<User> users = await this.database.Users
+            .Where(u => u.PermissionLevel != PermissionLevel.Banned && u.Username.Contains(query))
```
Slendy, line 66: How does the .Contains() here play with uppercase/lowercase, is this a concern or is it expected that you should capitalize the username properly?

Slendy (changes requested):
Few small concerns over formatting and username letter case. Other than that, this looks good to me. You could potentially add pagination but it's not needed for your current use case and most searches will probably return less than 20 results anyways. If you do decide to implement pagination at some point use keyset pagination rather than .Skip()

ProjectLighthouse.Servers.API/Controllers/UserEndpoints.cs
Zaprit: maybe, though I wasn't sure if that would clash with the endpoint that's already bound to /user/{id}

ProjectLighthouse.Servers.API/Controllers/UserEndpoints.cs
Zaprit: capitalisation seems to be ignored

ProjectLighthouse.Servers.API/Controllers/UserEndpoints.cs
Slendy: maybe it should be under a /search endpoint then, like `/api/v1/search/user?query=`?

ProjectLighthouse.Servers.API/Controllers/UserEndpoints.cs
Slendy: Should be fine then

Slendy (approved):
LGTM

PR #689, opened 2023-02-26
Allow environment variable expansion in config files

Slendy:
Now you can use environment variables in string variables for any config. The format is similar to windows because of how C# implements it. If you had a variable named `DB_ADDRESS` you would access it with `%DB_ADDRESS%`. I'm contemplating changing the default config values to include environment variables for easy setup on docker but I think that's outside the scope of this PR.


Zaprit (approved):
looks good to me

PR #702, opened 2023-03-10
Refactor serialization system

Slendy:
This PR does away with the whole LbpSerializer string concatenation serialization system and moves to a modified version of the C# XML serializer. It had to be changed because LBP doesn't exactly follow XML specifications, so things like namespaces and type-identifying attributes aren't used and must be removed from the normal output. There are some potential performance benefits to this PR because data for an object needed for serialization is now loaded separately rather than having to inner join an entire user with a slot just to get their username.
The main takeaway from this PR is that all database entities have been renamed to follow the <Name>Entity naming convention to differentiate them from other types. They have also been stripped of all serialization-related types, attributes, and methods so that they are strictly for storing data. Then there are objects that implement ILbpSerializable that have methods to take an Entity and convert it to a serializable class. For deserializing objects, there are methods in the serialization classes to convert an ILbpSerializable to its Entity counterpart.


W0lf4llo (commented):
I just quickly skimmed over most of the changes and from what I can see there doesn’t seem to be anything gamebreaking. 

TL;DR
LGTM

sudokoko (approved):
Not seeing anything production-breaking, though might be worth testing on staging.slendy.pw if not already.

LGTM.

daggintosh:
i declare this pr yolo'd

PR #712, opened 2023-03-16
Make in-game email system not lock after one attempt

Slendy:
I thought I already made this change since I implemented the `SendVerificationEmail` method but I guess I forgot.


W0lf4llo:
LGTM

PR #717, opened 2023-03-23
Allow head requests to status endpoints

Slendy:
Currently trying to send an HTTP HEAD request to any of the status endpoints you get a 405 Method not allowed. Some automated status checks use this method because it's lightweight and doesn't return a body and may interpret the 405 response as an error.


daggintosh (approved):
acceptverb to me

sudokoko (approved):
LGTM

image

PR #727, opened 2023-03-30
Fix doubly sanitized strings

Slendy:
After the recent serialization refactor, sanitization and desensitization are both handled by the built-in .NET systems. Before, only sanitization was handled and the manual string serialization would result in potential XML injection. Since this issue is no longer present the manually sanitized strings can be manually decoded.

### DIFF
--- a/ProjectLighthouse.Servers.Website/Controllers/SlotPageController.cs
+++ b/ProjectLighthouse.Servers.Website/Controllers/SlotPageController.cs
@@ -68,7 +68,6 @@ public class SlotPageController : ControllerBase
             return this.Redirect("~/slot/" + id);
         }
 
-        // Prevent potential xml injection and censor content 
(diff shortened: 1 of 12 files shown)


ProjectLighthouse.Servers.Website/Controllers/SlotPageController.cs
```diff
@@ -69,7 +69,6 @@ public async Task<IActionResult> PostComment([FromRoute] int id, [FromForm] stri
         }
 
         // Prevent potential xml injection and censor content
```
sudokoko, line 71:
```suggestion
```
Remove unnecessary comment

Slendy:
Tested the migration and it seems to work fine

## Before:
image
## After:
image

sudokoko (approved):
LGTM

PR #731, opened 2023-04-02
Fix memory leak in GameServer

Slendy:
Currently, there is a memory leak issue in the lighthouse GameServer. I believe it is caused by the allocation of ILbpSerializable objects before the final aggregation in LINQ. For example, to get a list of slots the current way is done like so:
```csharp
await this.database.Slots.Where(s => s.TeamPick).Take(30).Select(s => SlotBase.CreateFromEntity(s)).ToListAsync()
//                                                  ^ this is the problem
```
This results in SlotEntity not being garbage collected because somehow a reference is still held
My proposed fix is first to aggregate the SlotEntity query into a list and then call Select() and ToList(), which would look something like this:
```csharp
(await this.database.Slots.Where(s => s.TeamPick).Take(30).ToListAsync()).Select(s => SlotBase.CreateFromEntity(s)).ToList()
```

### DIFF
--- a/ProjectLighthouse.Servers.API/Controllers/SlotEndpoints.cs
+++ b/ProjectLighthouse.Servers.API/Controllers/SlotEndpoints.cs
@@ -1,6 +1,7 @@
 #nullable enable
 using LBPUnion.ProjectLighthouse.Configuration;
 using LBPUnion.ProjectLighthouse.Database;
+using LBPUnion.ProjectLighthouse.Extensions;
 using LBPUnion.ProjectLighthouse.Servers.API.Responses;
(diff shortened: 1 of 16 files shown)


daggintosh (approved):
Just don't break it

sudokoko (approved):
LGTM

Zaprit (approved):
Looks good, out of curiosity is there any performance penalty in doing it this was around compared to the original way, other than the obvious reduction in RAM.

Slendy:
> Looks good, out of curiosity is there any performance penalty in doing it this was around compared to the original way, other than the obvious reduction in RAM.

By this way do you mean the caching of XmlSerializer instances or do you mean changing the order of ToList() and Select()

PR #738, opened 2023-04-09
Display a lock icon next to locked levels

sudokoko:
Display a lock icon next to an `InitiallyLocked` level

image

### DIFF
--- a/ProjectLighthouse.Servers.Website/Extensions/FormattingExtensions.cs
+++ b/ProjectLighthouse.Servers.Website/Extensions/FormattingExtensions.cs
@@ -5,14 +5,7 @@ namespace LBPUnion.ProjectLighthouse.Servers.Website.Extensions;
 
 public static class FormattingExtensions
 {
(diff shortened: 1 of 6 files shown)


ProjectLighthouse.Servers.Website/Extensions/FormattingExtensions.cs
```diff
@@ -0,0 +1,26 @@
...
+public static class FormattingExtensions
+{
+    public static string GetLevelLockIcon(this SlotEntity slot)
+    {
+        return slot switch
+        {
+            { InitiallyLocked: true } => "icon lock",
+            _ => ""
+        };
+    }
```
Slendy, line 15:
```suggestion
    public static string GetLevelLockIcon(this SlotEntity slot) => slot.InitiallyLocked ? "icon lock" : "";
```

Slendy (changes requested):
GetLevelLockIcon can be simplified but that's about it

Slendy (approved):
LGTM

PR #745, opened 2023-04-18
Allow streaming chunk filetype in mod filter

uhwot:
Fixes publishing LBP3 levels that use dynamic thermometers.


Zaprit (approved):
Looks good. As long as it's been tested and works.

Slendy (approved):
LGTM

PR #754, opened 2023-04-26
Rewrite captcha system

Slendy:
This PR makes the captcha system use a DI-injected HttpClient which should help prevent issues where one request timing out prevents any more requests from being made. The timeout has also been adjusted to be 5 seconds rather than the default of 100 seconds. Many unused directives have also been removed by Rider Auto-Analysis.

### DIFF
--- a/ProjectLighthouse.Servers.Website/Startup/WebsiteStartup.cs
+++ b/ProjectLighthouse.Servers.Website/Startup/WebsiteStartup.cs
@@ -1,8 +1,11 @@
 using System.Globalization;
 using System.Net;
+using LBPUnion.ProjectLighthouse.Configuration;
+using LBPUnion.ProjectLighthouse.Configuration.ConfigurationCategories;
 using LBPUnion.ProjectLighthouse.Database;
 using LBPUnion.ProjectLighthouse.Localization;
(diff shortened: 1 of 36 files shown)


ProjectLighthouse.Servers.Website/Startup/WebsiteStartup.cs
```diff
@@ -44,6 +47,20 @@ public void ConfigureServices(IServiceCollection services)
 
         services.AddDbContext<DatabaseContext>();
 
+        services.AddHttpClient<ICaptchaService, CaptchaService>("CaptchaAPI",
+            client =>
+            {
+                Uri captchaUri = ServerConfiguration.Instance.Captcha.Type switch
+                {
+                    CaptchaType.HCaptcha => new Uri("https://hcaptcha.com"),
```
sudokoko, line 55: Is the hCaptcha API root URL correct? I don't use hCaptcha but I just wanted to make sure.

sudokoko: Just looked into the docs a bit, seems OK.

sudokoko (approved):
LGTM

PR #755, opened 2023-04-29
String fix with login/register button

sudokoko:
Only display "Login" as opposed to "Login / Register" when `ServerConfiguration.Instance.Authentication.RegistrationEnabled` is false

### DIFF
--- a/ProjectLighthouse.Localization/BaseLayout.resx
+++ b/ProjectLighthouse.Localization/BaseLayout.resx
@@ -24,11 +24,6 @@
         </value>
     </resheader>
 
-    <data name="header_home" xml:space="preserve">
-        <value>Home</value>
-        <comment>A button on the header that takes you to the landing page.</comment>
-    </data>
-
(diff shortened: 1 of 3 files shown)


ProjectLighthouse.Localization/BaseLayout.resx
```diff
@@ -49,6 +49,10 @@
         <comment>A button on the header that takes you to a list of authentication attempts.</comment>
     </data>
 
+    <data name="header_login" xml:space="preserve">
+        <value>Login</value>
+        <comment>A button on the header that lets you log in.</comment>
+    </data>
     <data name="header_loginRegister" xml:space="preserve">
```
Slendy, line 56: Since you're removing the header_home and header_profile TranslatableStrings the respective resource sections should be deleted too

Slendy (changes requested):
Looks mostly fine to me, just one small thing

Slendy (approved):
lgtm

PR #761, opened 2023-05-04
Add moderation/permission info to /api/v1/user

sudokoko:
Adds the permission level key into the response of `/api/v1/user`.


Slendy (changes requested):
This is the same thing 5 times, just include the user’s permissionlevel

Slendy (approved):

PR #773, opened 2023-05-26
Add proper ban page when logging in

sudokoko:
This PR reworks the existing banned user message upon logging in.

* Adds a separate page that displays details of your moderation case, including creation/expiration time upon logging in
* Adds translatable strings for the ban page
* Moves the banned check within the login page backend to allow for the page to correctly display

**Example:**
Screen Shot 2023-05-26 at 6 53 42 PM

### DIFF
--- a/ProjectLighthouse.Localization/Moderation.resx
+++ b/ProjectLighthouse.Localization/Moderation.resx
@@ -24,8 +24,8 @@
     <data name="ban_explanation" xml:space="preserve">
         <value>Your account has been banned due to violations of the {0} Terms of Service.</value>
     </data>
-    <data name="ban_created" xml:space="preserve">
-        <value>Ban Created: {0}</value>
+    <data name="ban_issued" xml:space="preserve">
+        <value>Ban Issued: {0}</value>
     </data>
     <data name="ban_expires" xml:space="preserve">
         <value>Ban Expires: {0}</value>
--- a/ProjectLighthouse.Localization/Moderation.resx
+++ b/ProjectLighthouse.Localization/Moderation.resx
@@ -18,22 +18,22 @@
     <resheader name="writer">
         <value>System.Resources.ResXResourceWriter, System.Windows.Forms, Version=2.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089</value>
     </resheader>
-    <data name="banned_heading" xml:space="preserve">
+    <data name="ban_heading" xml:space="preserve">
         <value>Account Banned</value>
     </data>
-    <data name="banned_explain" xml:space="preserve">
-        <value>Your account has been banned due to inappropriate, offensive, or violent behavior.</value>
+    <data name="ban_explanation" xml:space="preserve">
+        <value>Your account has been banned due to violations of the {0} Terms of Service.</value>
     </data>
-    <data name="case_created" xml:space="preserve">
-        <value>Case Created</value>
+    <data name="ban_created" xml:space="preserve">
+        <value>Ban Created: {0}</value>
     </data>
-    <data name="case_expires" xml:space="preserve">
-        <value>Case Expires</value>
+    <data name="ban_expires" xml:space="preserve">
+        <value>Ban Expires: {0}</value>
     </data>
     <data name="ban_reason" xml:space="preserve">
-        <value>Ban Reason</value>
+        <value>Ban Reason: {0}</value>
     </data>
-    <data name="appeal_instructions" xml:space="preserve">
-        <value>If you believe your ban was unjust, please contact an instance moderator.</value>
+    <data name="ban_does_not_expire" xml:space="preserve">
+        <value>Does not expire</value>
     </data>
 </root>
--- a/ProjectLighthouse.Localization/Moderation.resx
+++ b/ProjectLighthouse.Localization/Moderation.resx
@@ -18,22 +18,34 @@
     <resheader name="writer">
         <value>System.Resources.ResXResourceWriter, System.Windows.Forms, Version=2.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089</value>
     </resheader>
-    <data name="ban_heading" xml:space="preserve">
-        <value>Account Banned</value>
+    <data name="suspension_heading" xml:space="preserve">
+        <value>Account Suspended</value>
     </data>
-    <data name="ban_explanation" xml:space="preserve">
-        <value>Your account has been banned due to violations of the {0} Terms of Service.</value>
+    <data name="suspension_explanation" xml:space="preserve">
+        <value>Your {0} account has been suspended due to code of conduct violations.</value>
     </data>
-    <data name="ban_issued" xml:space="preserve">
-        <value>Ban Issued: {0}</value>
+    <data name="suspension_expiration" xml:space="preserve">
+        <value>During this suspension, the following features will be limited until {0}:</value>
     </data>
-    <data name="ban_expires" xml:space="preserve">
-        <value>Ban Expires: {0}</value>
+    <data name="suspension_reason" xml:space="preserve">
+        <value>Reason for suspension: "{0}"</value>
     </data>
-    <data name="ban_reason" xml:space="preserve">
-        <value>Ban Reason: {0}</value>
+    <data name="suspension_circumvent_warning" xml:space="preserve">
+        <value>Attempts to circumvent this suspension will result in an extended suspension period.</value>
     </data>
-    <data name="ban_does_not_expire" xml:space="preserve">
+    <data name="lbp_online_multiplayer" xml:space="preserve">
+        <value>LittleBigPlanet™ Online multiplayer</value>
+    </data>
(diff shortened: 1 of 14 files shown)


ProjectLighthouse.Servers.Website/Pages/Login/BannedUserPage.cshtml
```diff
@@ -0,0 +1,32 @@
...
+    TimeZoneInfo timeZoneInfo = TimeZoneInfo.FindSystemTimeZoneById(timeZone);
+}
+
+<div class="ui middle aligned left aligned">
+    <h1>@Model.Translate(ModerationStrings.BannedHeading)</h1>
+    <p>@Model.Translate(ModerationStrings.BannedExplain)</p>
+    <p>
+        <strong>@Model.Translate(ModerationStrings.CaseCreated):</strong>
+        @TimeZoneInfo.ConvertTime(Model.ModCase.CreatedAt, timeZoneInfo).ToString("M/d/yyyy @ h:mm tt") <br />  
+        @if (Model.ModCase.ExpiresAt != null)
```
Slendy, line 19: There should probably be something to have the date in a sentence using formatters. If you put arguments in the .resx file like {0}, {1}, etc. You can then call translate with multiple arguments that will be injected into the translatable string.

ProjectLighthouse.Servers.Website/Pages/Login/BannedUserPage.cshtml
```diff
@@ -0,0 +1,32 @@
...
+        {
+            <strong>@Model.Translate(ModerationStrings.CaseExpires):</strong>
+            @TimeZoneInfo.ConvertTime(Model.ModCase.ExpiresAt.Value, timeZoneInfo).ToString("M/d/yyyy @ h:mm tt")
```
Slendy, line 22: Might be helpful to also have something for when a ban is indefinite

ProjectLighthouse.Servers.Website/Pages/Login/BannedUserPage.cshtml.cs
```diff
@@ -0,0 +1,36 @@
...
+
+    public ModerationCaseEntity ModCase = null!;
+
+    [UsedImplicitly]
+    public async Task<IActionResult> OnGet()
+    {
+        UserEntity? user = this.Database.UserFromWebRequest(this.Request);
+
+        if (user == null) return this.Redirect("~/login");
+        if (user.IsBanned == false) return this.Redirect("~/");
```
Slendy, line 25:
```suggestion
        if (!user.IsBanned) return this.Redirect("~/");
```

ProjectLighthouse.Servers.Website/Pages/Login/LoginForm.cshtml.cs
```diff
@@ -120,6 +113,13 @@ public async Task<IActionResult> OnPost(string username, string password, string
                 : this.Redirect("~/2fa" + "?redirect=" + HttpUtility.UrlEncode(redirect));
         }
 
+        if (user.IsBanned)
+        {
+            Logger.Warn($"User {user.Username} (id: {user.UserId}) logged in, but is banned",
```
Slendy, line 118: The log message might be a little unnecessary now

ProjectLighthouse.Servers.Website/Pages/Login/BannedUserPage.cshtml
```diff
@@ -0,0 +1,32 @@
+﻿@page "/banned"
+@using LBPUnion.ProjectLighthouse.Localization.StringLists
+@model LBPUnion.ProjectLighthouse.Servers.Website.Pages.Login.BannedUserPage
+
+@{
+    Layout = null;
```
Slendy, line 6: Is there a reason this isn't `Layouts/BaseLayout`?

ProjectLighthouse.Localization/StringLists/ModerationStrings.cs
```diff
@@ -0,0 +1,16 @@
+﻿namespace LBPUnion.ProjectLighthouse.Localization.StringLists;
+
+public static class ModerationStrings
+{
+    public static readonly TranslatableString BannedHeading = create("banned_heading");
```
Slendy, line 5: The wording of some of these feels off. For these two you use banned as a verb then you use ban as a noun for the reason. I don't really mind which you use but I would keep them consistent.

ProjectLighthouse.Localization/StringLists/ModerationStrings.cs
```diff
@@ -0,0 +1,16 @@
...
+    public static readonly TranslatableString BannedExplain = create("banned_explain");
```
Slendy, line 6: This should probably be `_explaination` or `_description` instead of `_explain`

ProjectLighthouse.Localization/Moderation.resx
```diff
@@ -0,0 +1,39 @@
...
+        <value>System.Resources.ResXResourceReader, System.Windows.Forms, Version=2.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089</value>
+    </resheader>
+    <resheader name="writer">
+        <value>System.Resources.ResXResourceWriter, System.Windows.Forms, Version=2.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089</value>
+    </resheader>
+    <data name="banned_heading" xml:space="preserve">
+        <value>Account Banned</value>
+    </data>
+    <data name="banned_explain" xml:space="preserve">
+        <value>Your account has been banned due to inappropriate, offensive, or violent behavior.</value>
```
Slendy, line 25: Probably better to say something more vague like `You have been banned for violating the rules in our terms of service.`

ProjectLighthouse.Localization/Moderation.resx
```diff
@@ -0,0 +1,39 @@
...
+    </data>
+    <data name="case_created" xml:space="preserve">
```
Slendy, line 27: I'm just going off how other services usually handle bans, but you usually aren't told when you were banned. Generally, they just have the reason and when it expires.

ProjectLighthouse.Localization/Moderation.resx
```diff
@@ -0,0 +1,39 @@
...
+        <value>Case Created</value>
+    </data>
+    <data name="case_expires" xml:space="preserve">
+        <value>Case Expires</value>
+    </data>
+    <data name="ban_reason" xml:space="preserve">
+        <value>Ban Reason</value>
+    </data>
+    <data name="appeal_instructions" xml:space="preserve">
+        <value>If you believe your ban was unjust, please contact an instance moderator.</value>
```
Slendy, line 37: Unless we are also going to implement an appeals system at some point it seems better to exclude the hardcoded appeal. You could instead have moderators include instructions for appeal in the ban message itself.

Slendy (changes requested):
I think functionally everything looks good but I have some concerns over linguistics and formatting.

ProjectLighthouse.Localization/Moderation.resx
sudokoko: This is to keep consistent with what is displayed publicly on banned users profiles, but I'll take your word for it.

ProjectLighthouse.Localization/Moderation.resx
sudokoko: I'd have to check the current ban string but I believe it mentions something of the same aspect.

ProjectLighthouse.Localization/Moderation.resx
sudokoko: I'd have to check the current ban string but I believe it mentions something of the same aspect.

ProjectLighthouse.Localization/Moderation.resx
Metraberryy: You could change the public-facing message as well to keep it consistent, I feel it's too verbose anyway

ProjectLighthouse.Localization/Moderation.resx
sudokoko: I think it'd allow the user to have a general scope of their punishment length, opting to keep it.

ProjectLighthouse.Servers.Website/Pages/Login/BannedUserPage.cshtml
```diff
@@ -0,0 +1,49 @@
...
+            <span>
+                @Model.Translate(ModerationStrings.SuspensionExpiration, TimeZoneInfo.ConvertTime(Model.ModCase.ExpiresAt.Value, timeZoneInfo).ToString("M/d/yyyy @ h:mm tt"))
+            </span>
+        }
+        else
+        {
+            <span>
+                @Model.Translate(ModerationStrings.SuspensionExpiration, Model.Translate(ModerationStrings.DoesNotExpire))
+            </span>
+        }
```
Slendy, line 33: The span tags should probably be outside the if statement so you don't have to write an opening and closing tag for each if body

Slendy (changes requested):

Slendy (approved):

PR #777, opened 2023-05-31
Patch account suspended page display issues

sudokoko:
Patches two issues arising from #773.

* Two title headings would display on page
* Renderer would pull oldest case rather than newest, and would still show dismissed cases.
  * Technically, showing a dismissed case shouldn't happen anyways, but still a good thing to check for.


daggintosh (approved):
lgtm

PR #784, opened 2023-06-07
Refactor NpTicket parser and add NpTicket signature documentation

Slendy:
Makes ticket body parsing and ticket signature verification more abstract and modular rather than having everything done in a single file.


ProjectLighthouse.Tests/Unit/TicketTests.cs
```diff
@@ -0,0 +1,47 @@
...
+public class TicketTests
+{
+    [Fact]
+    public void CanReadTicket()
+    {
+        TicketBuilder builder = new TicketBuilder().SetCountry("br")
+            .SetUserId(21)
+            .SetDomain("us")
+            .SetStatus(0)
+            .SetIssuerId(0x74657374)
```
sudokoko, line 18:
```suggestion
    public void CanReadTicket()
    {
        TicketBuilder builder = new TicketBuilder()
            .SetCountry("br")
            .SetUserId(21)
            .SetDomain("us")
            .SetStatus(0)
            .SetIssuerId(0x74657374)
```
Can this be moved down to be consistent with styling?

sudokoko (approved):
Looks good functionally, Styling nitpick

ProjectLighthouse.Tests/Unit/TicketTests.cs
Slendy: I can, but any time the file is formatted Rider tries to move it back

sudokoko: Ah, that's alright then.

PR #789, opened 2023-06-11
Add online users endpoint to user API

sudokoko:
This PR adds an endpoint which lists all currently online users by their login time. This could be useful for services that want to deal with all currently online users at once but don't want to use some other inefficient method of obtaining the users.

### DIFF
--- a/ProjectLighthouse.Servers.API/Controllers/UserEndpoints.cs
+++ b/ProjectLighthouse.Servers.API/Controllers/UserEndpoints.cs
@@ -128,18 +128,23 @@ public class UserEndpoints : ApiEndpointController
     /// <summary>
     /// Gets a list of online users and returns it.
     /// </summary>
+    /// <param name="page">Page number. Increments skip value by clamp value</param>
+    /// <param name="clamp">Number of entries to clamp response to</param>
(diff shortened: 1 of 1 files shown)


Zaprit (approved):
LGTM

ProjectLighthouse.Servers.API/Controllers/UserEndpoints.cs
```diff
@@ -128,18 +128,23 @@ public async Task<IActionResult> CreateUserInviteToken([FromRoute] string? usern
...
+        if (clamp > 100) return this.BadRequest();
+
         List<ApiUser> onlineUsers = (await this.database.Users
             .Where(u => u.PermissionLevel != PermissionLevel.Banned)
             .Where(u => u.ProfileVisibility == PrivacyType.All)
             .Where(u => u.GetStatus(this.database).StatusType == StatusType.Online)
             .OrderBy(u => u.LastLogin)
-            .Take(50)
+            .Skip(page * clamp)
+            .Take(clamp)
```
sudokoko, line 147: This could probably be moved to it's own query extension in the future as well as cleaned up so it can be more easily used, but that's outside the scope of this PR.

PR #796, opened 2023-06-16
Refactor RepeatingTaskHandler

Slendy:
Refactors RepeatingTaskHandler into an ASP.NET service that uses proper scoped dependency injection instead of holding a manually instantiated DbContext.


sudokoko (approved):
LGTM

Zaprit:
May I ask what the point of this is/what it changes and why? Also should the website and the game server have the same repeating tasks, for instance should the website be cleaning up dive in rooms?

sudokoko:
> May I ask what the point of this is/what it changes and why? Also should the website and the game server have the same repeating tasks, for instance should the website be cleaning up dive in rooms?

It looks like this PR is simply refactoring the backbone of the repeating tasks handler. The actual tasks are not handled by the handler itself and have their own separate task files [here](https://github.com/LBPUnion/ProjectLighthouse/tree/bf2e36e8a484e263a1b457dafa472e002222c8ab/ProjectLighthouse/Administration/Maintenance/RepeatingTasks).

Zaprit:
> > May I ask what the point of this is/what it changes and why? Also should the website and the game server have the same repeating tasks, for instance should the website be cleaning up dive in rooms?
> 
> 
> 
> It looks like this PR is simply refactoring the backbone of the repeating tasks handler. The actual tasks are not "ran" in a sense by the handler itself and have their own separate task files [here](https://github.com/LBPUnion/ProjectLighthouse/tree/bf2e36e8a484e263a1b457dafa472e002222c8ab/ProjectLighthouse/Administration/Maintenance/RepeatingTasks).

I had seen that, though I'm wondering what the effects of this PR would be, also what on earth even is scoped dependency injection

sudokoko:
> > > May I ask what the point of this is/what it changes and why? Also should the website and the game server have the same repeating tasks, for instance should the website be cleaning up dive in rooms?
> > 
> > 
> > It looks like this PR is simply refactoring the backbone of the repeating tasks handler. The actual tasks are not "ran" in a sense by the handler itself and have their own separate task files [here](https://github.com/LBPUnion/ProjectLighthouse/tree/bf2e36e8a484e263a1b457dafa472e002222c8ab/ProjectLighthouse/Administration/Maintenance/RepeatingTasks).
> 
> I had seen that, though I'm wondering what the effects of this PR would be, also what on earth even is scoped dependency injection

Not quite sure, though the repeating tasks have had a few issues over time and perhaps this will solve it? (See third pinned down in R&D chat).

"Scoped dependency injection is a technique for creating objects that are the same within a request but different across different requests. Scoped objects are created once per request and are shared within a single request.
Scoped objects are different from transient and singleton objects. Transient objects are always different; a new instance is provided to every controller and every service. Singleton objects are the same for every object and every request." -ChatGPT

Slendy:
The current functionality is repeating tasks get run on the game server, the API, and the website. An ideal solution would be to integrate the tasks into the database so they are properly scheduled or to just use a library but IMO that's outside the scope of what this PR aims to fix. This PR aims to address the plethora of problems associated with the old implementation:
1. Has no support for cancellation or proper clean-up
2. Uses a long-living DbContext which is considered bad practice. DbContexts are very lightweight to create because they are designed to be instantiated and disposed for each web request.
3. It runs outside the scope of ASP.NET which doesn't really make sense when ASP.NET supports background services. (We are essentially trying to mimic a feature that already exists)
4. Uses busy waiting to wait for the next task to run.
5. The old implementation is static which is inherently untestable. It also has a hard dependency on MaintenanceHelper rather than taking a list of tasks as input.

Zaprit (approved):
Looks good to me, I understand what the point of this is now.

PR #795, opened 2023-06-14
Optimize censor filter system

sudokoko:
This PR makes several optimizations to the censor filtering system.
* Adds proper logging of censored messages
* Prevents garbage collection & memory problems with larger censored strings
* Dead code removal

PR #804, opened 2023-06-18
Censor filter logging improvements

sudokoko:
Self explanatory, piggyback off of #795.

Migrates configuration version from `v20` to `v21`.

Code has been tested to be working on a personal dev instance. Could be a bit neater but it shouldn't affect much.


daggintosh (approved):
image

I don't see anything particularly damaging to bacon and I trust that it does its job and doesn't break the moment someone writes the word "tomato" in a level comment

ProjectLighthouse/Helpers/CensorHelper.cs
```diff
@@ -47,13 +41,9 @@ public static string FilterMessage(string message)
...
             {
                 Censor(censorIndices[i], profanity.Length, stringBuilder);
-                profaneCount += 1;
             }
         }
-
-        if (ServerConfiguration.Instance.LogChatFiltering && profaneCount > 0 && message.Length <= lbpCharLimit)
-            Logger.Info($"Censored {profaneCount} profane words from message \"{message}\"", LogArea.Filter);
-
+
```
Slendy, line 46:
```suggestion
```

ProjectLighthouse.Servers.GameServer/Controllers/MessageController.cs
```diff
@@ -102,12 +105,15 @@ public async Task<IActionResult> Filter(IMailService mailService)
...
+        
+        string filteredText = CensorHelper.FilterMessage(message);
 
-        if (ServerConfiguration.Instance.LogChatFiltering)
-            Logger.Info($"{username}: {message} / {filteredText}", LogArea.Filter);
+        if (ServerConfiguration.Instance.LogChatFiltering && filteredText != message)
+            Logger.Info($"Censored profane word(s) from in-game text sent by {username}: \"{message}\" => \"{filteredText}\"",
+                LogArea.Filter);
+        if (ServerConfiguration.Instance.LogChatMessages)
+            Logger.Info($"{username}: \"{message}\"", LogArea.Filter);
```
Slendy, line 116: This should probably be above the FilterMessage call because
1. It doesn't use filteredText
2. If the FilterMessage call fails and throws an error it will never be logged and it would be harder to track down the root cause.

Slendy (changes requested):

Slendy (approved):
LGTM
image

PR #809, opened 2023-06-20
Allow moderators to moderate story levels

Slendy:
The primary reason for this PR is currently it isn't really possible for mods to delete abusive comments or scores on story levels.

This PR adds an extra comment on story levels that shows the story levels ID and slot ID. This comment only displays to moderators and will always be at the top of the comments. This is so that mods can easily find the story level id that a bad score/comment is on.
image

The mod panel on the website has a new form that lets people easily put in a story level ID and be redirected to the slot page.
image

Then finally, mods are now able to directly access story levels on the website, it looks something like this:
image


m88youngling:
A little hacky way of doing things, but looks like a good solution. Is there a reference table of level names or some way for moderators to know which story level they're looking at in a more familiar way than the dev slot ID?

sudokoko (approved):
As m88 said, a bit of a hacksaw solution but at least it gets the ball rolling for story level moderation. LGTM.

daggintosh (approved):
k

PR #824, opened 2023-06-30
Fix missing string on banned page when mod case is null

sudokoko:
Self explanatory.


speedyman09:
it looks like it works but i cant review it

sudokoko:
Closing this PR as I believe I have found the root cause of the edge case bug.

PR #825, opened 2023-06-30
Fix null case bug in BannedUserPage and properly add no expiration text

sudokoko:
This PR fixes the issue where the ban information page would improperly pull the user's latest moderation case, pulling the latest dismissed case rather than the latest case where expiration isn't null (which should theoretically never happen anyways, but should still check since it is marked as nullable in the database).

Also properly implements the existing permanent ban string into the information page, as well as fixes a missing page element when the case is null.

### DIFF
--- a/ProjectLighthouse.Servers.Website/Pages/Login/BannedUserPage.cshtml
+++ b/ProjectLighthouse.Servers.Website/Pages/Login/BannedUserPage.cshtml
@@ -9,6 +9,8 @@
 
     string timeZone = (string?)ViewData["TimeZone"] ?? TimeZoneInfo.Local.Id;
     TimeZoneInfo timeZoneInfo = TimeZoneInfo.FindSystemTimeZoneById(timeZone);
+
+    DateTime maximumExpiration = new(9999, 12, 31, 11, 59, 00, DateTimeKind.Utc);
 }
 
 <div class="ui middle aligned left aligned">
@@ -20,7 +22,7 @@
         @if (Model.ModCase != null)
         {
             <span>
-                @if (Model.ModCase.ExpiresAt != null)
+                @if (Model.ModCase.ExpiresAt < maximumExpiration)
                 {
                     @Model.Translate(ModerationStrings.SuspensionExpiration, TimeZoneInfo.ConvertTime(Model.ModCase.ExpiresAt.Value, timeZoneInfo).ToString("M/d/yyyy @ h:mm tt"))
                 }
@@ -29,6 +31,12 @@
                     @Model.Translate(ModerationStrings.SuspensionExpiration, Model.Translate(ModerationStrings.DoesNotExpire))
                 }
             </span>    
+        }
+        else
+        {
+            <span>
+                @Model.Translate(ModerationStrings.SuspensionExpiration, Model.Translate(GeneralStrings.Unknown))
+            </span>
         }
 
         <ul>
--- a/ProjectLighthouse.Servers.Website/Pages/Login/BannedUserPage.cshtml
+++ b/ProjectLighthouse.Servers.Website/Pages/Login/BannedUserPage.cshtml
@@ -1,6 +1,7 @@
 ﻿@page "/banned"
 @using LBPUnion.ProjectLighthouse.Configuration
 @using LBPUnion.ProjectLighthouse.Localization.StringLists
+@using LBPUnion.ProjectLighthouse.Servers.Website.Pages.Login
 @model LBPUnion.ProjectLighthouse.Servers.Website.Pages.Login.BannedUserPage
 
 @{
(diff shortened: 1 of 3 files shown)


ProjectLighthouse.Servers.Website/Pages/Login/BannedUserPage.cshtml
```diff
@@ -20,7 +22,7 @@
         @if (Model.ModCase != null)
         {
             <span>
-                @if (Model.ModCase.ExpiresAt != null)
+                @if (Model.ModCase.ExpiresAt < maximumExpiration)
```
Slendy, line 25:
```suggestion
                @if (Model.ModCase.ExpiresAt < DateTime.MaxValue)
```

ProjectLighthouse.Servers.Website/Pages/Login/BannedUserPage.cshtml
```diff
@@ -9,6 +9,8 @@
 
     string timeZone = (string?)ViewData["TimeZone"] ?? TimeZoneInfo.Local.Id;
     TimeZoneInfo timeZoneInfo = TimeZoneInfo.FindSystemTimeZoneById(timeZone);
+
+    DateTime maximumExpiration = new(9999, 12, 31, 23, 59, 00, DateTimeKind.Utc);
```
Slendy, line 13:
```suggestion
```

ProjectLighthouse.Servers.Website/Pages/Login/BannedUserPage.cshtml
```diff
@@ -30,6 +32,12 @@
...
             </span>    
         }
+        else
+        {
+            <span>
+                @Model.Translate(ModerationStrings.SuspensionExpiration, Model.Translate(GeneralStrings.Unknown))
+            </span>
+        }
 
         <ul>
```
Slendy, line 42: According to HTML specs and Rider analysis, you shouldn't be putting a list inside a paragraph object, you can either move the list outside the paragraph tag or turn it into a div

ProjectLighthouse.Servers.Website/Pages/Login/BannedUserPage.cshtml
```diff
@@ -30,6 +32,12 @@
...
             <li>@Model.Translate(ModerationStrings.LbpOnlineMultiplayer)</li>
```
Slendy, line 43: GitHub doesn't let me comment on other lines but on line 53 there is no handler for if a reason isn't given, it just looks like `Reason for suspension: ""`

ProjectLighthouse.Servers.Website/Pages/Login/BannedUserPage.cshtml
```diff
@@ -20,7 +22,7 @@
         @if (Model.ModCase != null)
         {
             <span>
-                @if (Model.ModCase.ExpiresAt != null)
+                @if (Model.ModCase.ExpiresAt < maximumExpiration)
                 {
                     @Model.Translate(ModerationStrings.SuspensionExpiration, TimeZoneInfo.ConvertTime(Model.ModCase.ExpiresAt.Value, timeZoneInfo).ToString("M/d/yyyy @ h:mm tt"))
                 }
```
Slendy, line 28: On line 31, the wording sounds odd in the full sentence:
```
During this suspension, the following features will be limited until Does not expire:
```

Slendy (changes requested):

ProjectLighthouse.Servers.Website/Pages/Login/BannedUserPage.cshtml
sudokoko: This is intentional. The MaxValue expression actually exceeds the maximum value the form lets you submit.

Slendy: Is it possible to make this field static then so it isn't being recreated for every request?

sudokoko: Will do that once I become available

ProjectLighthouse.Servers.Website/Pages/Login/BannedUserPage.cshtml.cs
```diff
@@ -34,7 +34,8 @@ public async Task<IActionResult> OnGet()
         if (user == null) return this.Redirect("~/login");
         if (!user.IsBanned) return this.Redirect("~/");
 
-        ModerationCaseEntity? modCase = await this.Database.Cases.OrderByDescending(c => c.CreatedAt)
+        ModerationCaseEntity? modCase = await this.Database.Cases
+			.OrderByDescending(c => c.CreatedAt)
```
sudokoko, line 38: Will fix this momentarily, Rider being weird.

Slendy (approved):
image

daggintosh (approved):

PR #830, opened 2023-07-06
Migrate scores to use proper relationships

Slendy:
The old way scores worked was using players' usernames in a comma-delimited list which has obvious problems like being able to spoof scores because the system ultimately trusts the score uploader. Other problems include different permutations of usernames will create entirely new scores instead of updating old ones.

The migration uses raw SQL because after the migration the entity is no longer compatible and would be complicated to implement with EF as it is now. There is a new feature in NET 8.0 to map raw SQL queries to objects but it doesn't release until November of this year.

Some drawbacks of the new system are that non-existent users are no longer allowed to keep scores. This also means that once a user is deleted, all of their scores will also be deleted. In some cases this is beneficial; for example, if a single user cheated many scores you can remove all of their scores by simply deleting their account.

Since this PR also creates more scores per level since every user gets their own score regardless of the number of players that were in the session, there also needs to be a rework of the score viewer on the website to have pagination and filter by score type but that's outside the scope of this PR.

# TODO 
- [x] Test migration more (maybe write unit tests)
- [x] Test in-game uploading and reading of scores in different games
- [x] Change website leaderboard to handle ties like the games (implemented in #839)

In some cases where people have multiple scores under the same level and category will have their highest score selected and kept.


ProjectLighthouse.Servers.GameServer/Controllers/Resources/ResourcesController.cs
```diff
@@ -42,11 +42,11 @@ public IActionResult GetResource(string hash)
         string fullPath = Path.GetFullPath(path);
 
         // Prevent directory traversal attacks
-        if (!fullPath.StartsWith(FileHelper.FullResourcePath)) return this.BadRequest();
+        if (!fullPath.StartsWith(FileHelper.FullResourcePath)) return this.StatusCode(400);
 
         if (FileHelper.ResourceExists(hash)) return this.File(IOFile.OpenRead(path), "application/octet-stream");
 
-        return this.NotFound();
+        return this.StatusCode(404);
```
sudokoko, line 49: Reason for these? Just curious.

ProjectLighthouse/Types/Entities/Level/ScoreEntity.cs
```diff
@@ -1,32 +1,29 @@
...
+using LBPUnion.ProjectLighthouse.Types.Entities.Profile;
 
 namespace LBPUnion.ProjectLighthouse.Types.Entities.Level;
 
 public class ScoreEntity
 {
     [Key]
     public int ScoreId { get; set; }
-
+
```
sudokoko, line 11:
```suggestion

```

ProjectLighthouse/Types/Entities/Level/ScoreEntity.cs
```diff
@@ -1,32 +1,29 @@
...
-    [XmlIgnore]
+    
     [ForeignKey(nameof(SlotId))]
     public SlotEntity Slot { get; set; }
-
-    [XmlIgnore]
+    
     public int ChildSlotId { get; set; }
-
+
```
sudokoko, line 18:
```suggestion

```

sudokoko (approved):
LGTM, just a few nitpicks and questions

ProjectLighthouse.Servers.GameServer/Controllers/Resources/ResourcesController.cs
Zaprit: I agree with koko, did these not work before

Zaprit (approved):
Looks good, see my question in the comment though

sudokoko (approved):

sudokoko (approved):
Decided to take another look at this since there have been changes since I approved a month ago. Still looks good.

PR #831, opened 2023-07-07
Add the ability for moderators to forcibly lock levels

sudokoko:
This PR adds a new mod case type which allows moderators to forcibly lock levels. Forcibly locking a level sets the level's InitiallyLocked variable, as well as the reason for the lock. It also sets the LockedByModerator variable, which is checked upon re-publish to make sure the user cannot just unlock the level.

<details>
<summary>Images</summary>
<br>

image
image

</details>

### DIFF
--- a/ProjectLighthouse.Servers.Website/Pages/SlotPage.cshtml
+++ b/ProjectLighthouse.Servers.Website/Pages/SlotPage.cshtml
@@ -26,12 +26,40 @@
 @if (Model.Slot!.Hidden)
 {
     <div class="ui inverted red segment">
-        <h2>This level is currently hidden.</h2>
-        <p><b>Only you and moderators may view this level.</b></p>
-        
-        <b>Reason:</b> <span>"@Model.Slot.HiddenReason"</span>
-        
-        <p><b>For more information please contact a moderator.</b></p>
+        <h3 style="margin-bottom:5px;"><i class="ban icon"></i> This level is currently hidden.</h3>
+        @if (Model.User != null && Model.User.IsModerator)
+        {
+            <b>Reason:</b>
+            <span>"@Model.Slot.HiddenReason"</span>
+            <br/>
+            <p>
+                <i>Only you and other moderators may view the hidden reason.</i>
+            </p>
+        }
+        else
+        {
+            <p>This level has been hidden for violating the Terms of Service. Remember to follow the rules!</p>
+        }
+    </div>
+}
+
+@if (Model.Slot!.LockedByModerator)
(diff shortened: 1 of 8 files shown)


Zaprit (approved):
LGTM

ProjectLighthouse.Servers.Website/Pages/SlotPage.cshtml
```diff
@@ -26,12 +26,40 @@
 @if (Model.Slot!.Hidden)
 {
     <div class="ui inverted red segment">
-        <h2>This level is currently hidden.</h2>
-        <p><b>Only you and moderators may view this level.</b></p>
-        
-        <b>Reason:</b> <span>"@Model.Slot.HiddenReason"</span>
-        
-        <p><b>For more information please contact a moderator.</b></p>
+        <h3 style="margin-bottom:5px;"><i class="ban icon"></i> This level is currently hidden.</h3>
```
Slendy, line 29: Is the inline style necessary on the header? Could you move the icon instead or does that look weird?

ProjectLighthouse/ProjectLighthouse/Migrations/DatabaseModelSnapshot.cs
```diff
@@ -406,6 +406,13 @@ protected override void BuildModel(ModelBuilder modelBuilder)
                     b.Property<ulong>("LocationPacked")
                         .HasColumnType("bigint unsigned");
 
+                    b.Property<bool>("LockedByModerator")
+                        .HasColumnType("tinyint(1)");
+
+                    b.Property<string>("LockedReason")
+                        .IsRequired()
```
Slendy, line 413: LockedReason is required but has no default value for previous records. This seems like it might cause issues, have you already tested this migration?

Slendy (changes requested):
Looks fine, I just have some questions about styling and the migration

ProjectLighthouse/ProjectLighthouse/Migrations/DatabaseModelSnapshot.cs
sudokoko: I had no issues when migrating a copy of the live Beacon database.

ProjectLighthouse.Servers.Website/Pages/SlotPage.cshtml
sudokoko: Yeah, give me a moment.

Slendy (approved):
lgtm

PR #838, opened 2023-07-14
Add proper relationships to Comments

Slendy:
Instead of having a TargetId that can represent either a UserId or a SlotId, I separated them into nullable TargetUserId and TargetSlotId columns. This has the benefits of indexing, and cascading deletes so that when a user or level is deleted all of the comments are also deleted with it.


ProjectLighthouse/Types/Entities/Profile/CommentEntity.cs
```diff
@@ -33,30 +46,23 @@ public class CommentEntity
...
-        {
+        if (deletedBy == null) return "This comment has been deleted";
+
+        // If the owner of the comment section deletes
+        if (deletedBy.UserId == this.TargetUserId || deletedBy.UserId == database.Slots.Find(this.TargetSlotId)?.CreatorId)
             return "This comment has been deleted by the player.";
-        }
 
-        return "This comment has been deleted.";
+        return "This comment has been deleted";
```
sudokoko, line 66: Is this how it appears in game?

sudokoko (approved):
LGTM, see comment

W0lf4llo:
LGTM

ProjectLighthouse/Migrations/20230714212153_AddUserIdAndSlotIdToComment.cs
```diff
@@ -0,0 +1,46 @@
...
+                table: "Comments",
+                type: "int",
+                nullable: true);
+
+            // Set SlotId and UserId to null if there isn't an existing slot or user
+            migrationBuilder.Sql("UPDATE Comments AS c INNER JOIN Users AS u ON c.TargetId = u.UserId AND c.Type = 0 SET c.TargetUserId = u.UserId");
+            migrationBuilder.Sql("UPDATE Comments AS c INNER JOIN Slots AS s ON c.TargetId = s.SlotId AND c.Type = 1 SET c.TargetSlotId = s.SlotId");
+
+            // Delete rows that have null SlotIds or UserIds
+            migrationBuilder.Sql("DELETE FROM Comments WHERE TargetUserId IS NULL AND TargetSlotId IS NULL");
```
sudokoko, line 32: Would this be beneficial to implement as an admin maintenance feature, sort of how you can delete broken photos? This can be applied to other areas as well.

sudokoko (approved):
This still looks good, see comment

ProjectLighthouse/Migrations/20230714212153_AddUserIdAndSlotIdToComment.cs
sudokoko: > This has the benefits of indexing, and cascading deletes so that when a user or level is deleted all of the comments are also deleted with it.

Unsure if above would defeat the purpose of this but might still be useful?

Zaprit (approved):
LGTM

ProjectLighthouse/Migrations/20230714212153_AddUserIdAndSlotIdToComment.cs
Slendy: By creating a foriegn key with a cascading delete, MySQL will automatically delete any comments that are on a user/level when it is deleted. Since there were no constraints before, some of the comments are on deleted levels or users and need to be manually deleted for the foreign key to be successfully created.

PR #841, opened 2023-07-15
Implement profile and level privacy settings

### LINKED ISSUE #568: Privacy Settings
The following privacy settings should be added to the website for users to configure:

- [x] Blocking other users
- [ ] Enabling/disabling profile comments
- [ ] Enabling/disabling level comments

There are also two in-game privacy settings we could choose to enable or disable as well. The options are to allow your levels and profile to be seen by everyone, only logged in users, or no one on the website. For public beta I'm not too concerned about these. Regardless of the user's choice, moderators should always be able to see the levels and profile on the website

sudokoko:
This PR implements a new settings area allowing you to set your privacy settings for your profile/levels. This recreates the functionality that LBP.me had, however these options would be set in game rather than through the website when LBP.me was a thing.

image

From the new settings page, you're able to **enable/disable your profile comments** (does not allow you to override an active UserDisableComments case), as well as **set your profile/level visibility settings**.

You're able to use the following three options for profile/level privacy;
* Public - shows your profile or levels to everyone
* Signed-In Only - only shows your profile or levels to users who are signed into the site
* In-Game Only - only shows your profile or levels to users who are in-game.

<details>
<summary>Further images</summary>
<br>

image
image
image
image

</details>

Closes #568.

### DIFF
--- a/ProjectLighthouse.Servers.Website/Pages/Partials/CommentsPartial.cshtml
+++ b/ProjectLighthouse.Servers.Website/Pages/Partials/CommentsPartial.cshtml
@@ -1,5 +1,4 @@
 ﻿@using System.Web
-@using System.IO
 @using LBPUnion.ProjectLighthouse.Database
 @using LBPUnion.ProjectLighthouse.Localization
 @using LBPUnion.ProjectLighthouse.Servers.Website.Extensions
@@ -17,11 +16,11 @@
 <div class="ui yellow segment" id="comments">
     @if (Model.Comments.Count == 0 && Model.CommentsEnabled)
     {
-        <p><i>There are no comments.</i></p>
+        <span><i>There are no comments.</i></span>
     }
     else if (!Model.CommentsEnabled)
     {
-        <p><i>Comments are disabled.</i></p>
+        <span><i>Comments are disabled.</i></span>
     }
     else
     {
--- a/ProjectLighthouse.Servers.Website/Pages/Partials/CommentsPartial.cshtml
+++ b/ProjectLighthouse.Servers.Website/Pages/Partials/CommentsPartial.cshtml
@@ -50,10 +50,9 @@
             CommentEntity comment = commentAndReaction.Key;
             int yourThumb = commentAndReaction.Value?.Rating ?? 0;
             DateTimeOffset timestamp = DateTimeOffset.FromUnixTimeSeconds(comment.Timestamp / 1000).ToLocalTime();
-            StringWriter messageWriter = new();
-            HttpUtility.HtmlDecode(comment.GetCommentMessage(Database), messageWriter);
 
-            string decodedMessage = messageWriter.ToString();
+            string decodedMessage = HttpUtility.HtmlDecode(comment.GetCommentMessage(Database));
+            
             string? url = Url.RouteUrl(ViewContext.RouteData.Values);
             if (url == null) continue;
 
--- a/ProjectLighthouse.Servers.Website/Pages/SlotPage.cshtml.cs
+++ b/ProjectLighthouse.Servers.Website/Pages/SlotPage.cshtml.cs
@@ -22,7 +22,7 @@ public class SlotPage : BaseLayout
     public bool CommentsEnabled;
     public readonly bool ReviewsEnabled = ServerConfiguration.Instance.UserGeneratedContentLimits.LevelReviewsEnabled;
 
-    public bool CanAccessSlot;
+    public bool CanViewSlot;
 
     public SlotEntity? Slot;
     public SlotPage(DatabaseContext database) : base(database)
@@ -36,10 +36,11 @@ public class SlotPage : BaseLayout
         if (slot == null) return this.NotFound();
         System.Diagnostics.Debug.Assert(slot.Creator != null);
 
+        bool isAuthenticated = this.User != null;
+        bool isOwner = slot.Creator == this.User || this.User != null && this.User.IsModerator;
+        
         // Determine if user can view slot according to creator's privacy settings
-        this.CanAccessSlot = slot.Creator.LevelVisibility.CanAccess(
-            this.User != null, 
-            slot.Creator == this.User || this.User != null && this.User.IsModerator);
+        this.CanViewSlot = slot.Creator.LevelVisibility.CanAccess(isAuthenticated, isOwner);
 
         if ((slot.Hidden || slot.SubLevel && (this.User == null && this.User != slot.Creator)) && !(this.User?.IsModerator ?? false))
             return this.NotFound();
--- a/ProjectLighthouse.Servers.Website/Pages/SlotPage.cshtml.cs
+++ b/ProjectLighthouse.Servers.Website/Pages/SlotPage.cshtml.cs
@@ -22,6 +22,8 @@ public class SlotPage : BaseLayout
     public bool CommentsEnabled;
     public readonly bool ReviewsEnabled = ServerConfiguration.Instance.UserGeneratedContentLimits.LevelReviewsEnabled;
 
+    public bool SlotPrivate;
+
     public SlotEntity? Slot;
     public SlotPage(DatabaseContext database) : base(database)
     {}
@@ -37,22 +39,28 @@ public class SlotPage : BaseLayout
         // Determine if user can view slot according to creator's privacy settings
         if (this.User == null || !this.User.IsAdmin)
         {
-            switch (slot.Creator.ProfileVisibility)
+            switch (slot.Creator.LevelVisibility)
             {
-                case PrivacyType.PSN:
-                {
-                    if (this.User != null) return this.NotFound();
-
-                    break;
-                }
                 case PrivacyType.Game:
                 {
-                    if (this.User == null || slot.Creator != this.User) return this.NotFound();
-
+                    if (this.User == null || slot.Creator != this.User) this.SlotPrivate = true;
                     break;
                 }
-                case PrivacyType.All: break;
-                default: throw new ArgumentOutOfRangeException();
+                case PrivacyType.PSN:
+                {
+                    if (this.User != null) this.SlotPrivate = true;
+                    break;
+                }
+                case PrivacyType.All:
+                {
+                    this.SlotPrivate = false;
+                    break;
+                }
+                default:
+                {
+                    this.SlotPrivate = false;
+                    break;
+                }
             }
         }
 
--- a/ProjectLighthouse.Servers.Website/Pages/SlotPage.cshtml.cs
+++ b/ProjectLighthouse.Servers.Website/Pages/SlotPage.cshtml.cs
@@ -48,7 +48,7 @@ public class SlotPage : BaseLayout
                 }
                 case PrivacyType.PSN:
                 {
-                    if (this.User != null) this.SlotPrivate = true;
+                    if (this.User == null) this.SlotPrivate = true;
                     break;
                 }
                 case PrivacyType.All:
--- a/ProjectLighthouse.Servers.Website/Pages/SlotPage.cshtml.cs
+++ b/ProjectLighthouse.Servers.Website/Pages/SlotPage.cshtml.cs
@@ -39,7 +39,7 @@ public class SlotPage : BaseLayout
         // Determine if user can view slot according to creator's privacy settings
         this.CanAccessSlot = slot.Creator.LevelVisibility.CanAccess(
             this.User != null, 
-            slot.Creator == this.User || !(this.User != null && this.User.IsModerator));
+            slot.Creator == this.User || this.User != null && this.User.IsModerator);
 
         if ((slot.Hidden || slot.SubLevel && (this.User == null && this.User != slot.Creator)) && !(this.User?.IsModerator ?? false))
             return this.NotFound();
--- a/ProjectLighthouse.Servers.Website/Pages/SlotPage.cshtml.cs
+++ b/ProjectLighthouse.Servers.Website/Pages/SlotPage.cshtml.cs
@@ -22,7 +22,7 @@ public class SlotPage : BaseLayout
     public bool CommentsEnabled;
     public readonly bool ReviewsEnabled = ServerConfiguration.Instance.UserGeneratedContentLimits.LevelReviewsEnabled;
 
-    public bool IsSlotPrivate;
+    public bool CanAccessSlot;
 
     public SlotEntity? Slot;
     public SlotPage(DatabaseContext database) : base(database)
@@ -37,7 +37,9 @@ public class SlotPage : BaseLayout
         System.Diagnostics.Debug.Assert(slot.Creator != null);
 
         // Determine if user can view slot according to creator's privacy settings
-        this.IsSlotPrivate = slot.Creator.LevelVisibility.IsPrivate(this.User != null, slot.Creator == this.User);
+        this.CanAccessSlot = slot.Creator.LevelVisibility.CanAccess(
+            this.User != null, 
+            slot.Creator == this.User || !(this.User != null && this.User.IsModerator));
 
         if ((slot.Hidden || slot.SubLevel && (this.User == null && this.User != slot.Creator)) && !(this.User?.IsModerator ?? false))
             return this.NotFound();
--- a/ProjectLighthouse.Servers.Website/Pages/SlotPage.cshtml.cs
+++ b/ProjectLighthouse.Servers.Website/Pages/SlotPage.cshtml.cs
@@ -22,7 +22,7 @@ public class SlotPage : BaseLayout
     public bool CommentsEnabled;
     public readonly bool ReviewsEnabled = ServerConfiguration.Instance.UserGeneratedContentLimits.LevelReviewsEnabled;
 
-    public bool SlotPrivate;
+    public bool IsSlotPrivate;
 
     public SlotEntity? Slot;
     public SlotPage(DatabaseContext database) : base(database)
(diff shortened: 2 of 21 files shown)


ProjectLighthouse.Servers.Website/Pages/Partials/CommentsPartial.cshtml
```diff
@@ -1,5 +1,4 @@
 ﻿@using System.Web
-@using System.IO
```
Zaprit, line 2: why did this even have system.io imported anyway????

ProjectLighthouse.Servers.Website/Pages/UserPage.cshtml
```diff
@@ -106,172 +113,199 @@
...
+        </div>
+        @if (isMobile)
+        {
+            <br/>
+        }    
     }
 </div>
 
-<div class="ui grid">
-    @{
```
Zaprit, line 139: probably out the scope of this PR but UI changes should NOT be done on the serverside, there's a whole language for this, it's called CSS (this is more as a general thing for lighthouse as a whole, but this here is a really good example of why)

ProjectLighthouse.Servers.Website/Pages/UserPage.cshtml.cs
```diff
@@ -18,44 +18,82 @@ public class UserPage : BaseLayout
...
+                    break;
+                }
+                case PrivacyType.All:
+                {
+                    this.LevelsPrivate = false;
+                    break;
+                }
+                default:
+                {
+                    this.LevelsPrivate = false;
```
Zaprit, line 94: This can be consolidated to save an if
```suggestion
        if (this.User == null || !this.User.IsModerator)
        {
            // Determine if user can view profile according to profileUser's privacy settings
            switch (this.ProfileUser.ProfileVisibility)
            {
                case PrivacyType.Game:
                {
                    if (this.ProfileUser != this.User) this.ProfilePrivate = true;
                    break;
                }
                case PrivacyType.PSN:
                {
                    if (this.User == null) this.ProfilePrivate = true;
                    break;
                }
                case PrivacyType.All:
                {
                    this.ProfilePrivate = false;
                    break;
                }
                default:
                {
                    this.ProfilePrivate = false;
                    break;
                }
            }

            // Determine if user can view levels according to profileUser's privacy settings
            switch (this.ProfileUser.LevelVisibility)
            {
                case PrivacyType.Game:
                {
                    if (this.ProfileUser != this.User) this.LevelsPrivate = true;
                    break;
                }
                case PrivacyType.PSN:
                {
                    if (this.User == null) this.LevelsPrivate = true;
                    break;
                }
                case PrivacyType.All:
                {
                    this.LevelsPrivate = false;
                    break;
                }
                default:
                {
                    this.LevelsPrivate = false;
                    break;
                }
            }
        }
```

ProjectLighthouse.Servers.Website/Pages/UsersPage.cshtml.cs
```diff
@@ -37,8 +37,9 @@ public async Task<IActionResult> OnGet([FromRoute] int pageNumber, [FromQuery] s
 
         if (this.PageNumber < 0 || this.PageNumber >= this.PageAmount) return this.Redirect($"/users/{Math.Clamp(this.PageNumber, 0, this.PageAmount - 1)}");
 
-        this.Users = await this.Database.Users.Where(u => u.PermissionLevel != PermissionLevel.Banned && u.Username.Contains(this.SearchValue))
-            .Where(u => u.ProfileVisibility == PrivacyType.All) // TODO: change check for when user is logged in
+        this.Users = await this.Database.Users
```
Zaprit, line 40: Should people with private profiles be included in this list?

Zaprit (changes requested):
See suggestions

ProjectLighthouse.Servers.Website/Pages/Partials/CommentsPartial.cshtml
sudokoko: Not sure. Rider removed this on it's own

ProjectLighthouse.Servers.Website/Pages/UserPage.cshtml
sudokoko: I believe slendy is working on a solution for this, not sure

ProjectLighthouse.Servers.Website/Pages/UsersPage.cshtml.cs
sudokoko: This was done on purpose - since you should still be able to search for a user, as the information that would be shown there is the same thing shown on the user page anyways regardless of privacy settings.

sudokoko (commented):
Responses to @Zaprit

ProjectLighthouse.Servers.Website/Pages/UserPage.cshtml.cs
```diff
@@ -18,44 +19,81 @@ public class UserPage : BaseLayout
...
+    { }
 
     public async Task<IActionResult> OnGet([FromRoute] int userId)
     {
         this.ProfileUser = await this.Database.Users.FirstOrDefaultAsync(u => u.UserId == userId);
         if (this.ProfileUser == null) return this.NotFound();
 
         // Determine if user can view profile according to profileUser's privacy settings
-        if (this.User == null || !this.User.IsAdmin)
+        if (this.User == null || !this.User.IsModerator)
```
Slendy, line 48: I don't know if this was intentional or not but on the UserPage you let mods bypass privacy settings but on the SlotPage you only let admins bypass the privacy settings

ProjectLighthouse.Servers.Website/Pages/UserPage.cshtml.cs
```diff
@@ -18,44 +19,81 @@ public class UserPage : BaseLayout
...
                 {
-                    if (this.ProfileUser != this.User) return this.NotFound();
+                    this.ProfilePrivate = false;
+                    break;
+                }
+                default:
+                {
+                    this.ProfilePrivate = false;
+                    break;
+                }
```
Slendy, line 71: This can be simplified into just the default block

ProjectLighthouse.Servers.Website/Pages/UserPage.cshtml.cs
```diff
@@ -18,44 +19,81 @@ public class UserPage : BaseLayout
...
+                case PrivacyType.All:
+                {
+                    this.SlotsPrivate = false;
+                    break;
+                }
+                default:
+                {
+                    this.SlotsPrivate = false;
                     break;
                 }
```
Slendy, line 96: Simplify
```suggestion
                default:
                {
                    this.SlotsPrivate = false;
                    break;
                }
```
Or if rider complains put both labels above the same block

ProjectLighthouse.Servers.Website/Pages/UserPage.cshtml.cs
```diff
@@ -116,7 +157,8 @@ public async Task<IActionResult> OnGet([FromRoute] int userId)
 
         foreach (KeyValuePair<CommentEntity, RatedCommentEntity?> kvp in this.Comments)
         {
-            RatedCommentEntity? reaction = await this.Database.RatedComments.Where(r => r.CommentId == kvp.Key.CommentId)
+            RatedCommentEntity? reaction = await this.Database.RatedComments
```
Slendy, line 160: I understand that you prefer this style of indentation and I'm open to having a conversation about changing the entire project to this style, but for now, instead of changing them one by one we should just use the old style and then at a later date we can reformat the entire project

ProjectLighthouse.Servers.Website/Pages/UserPrivacyPage.cshtml
```diff
@@ -0,0 +1,107 @@
...
+                    <select class="ui fluid dropdown" type="text" name="profilePrivacyLevel" id="profilePrivacyLevel">
+                        <option value="public" @(Model.ProfileUser.ProfileVisibility == PrivacyType.All ? "selected" : "")>
+                            Share your profile with everyone!
+                        </option>
+                        <option value="signedInOnly" @(Model.ProfileUser.ProfileVisibility == PrivacyType.PSN ? "selected" : "")>
+                            Only share your profile with users who are signed into the website or playing in-game.
+                        </option>
+                        <option value="inGameOnly" @(Model.ProfileUser.ProfileVisibility == PrivacyType.Game ? "selected" : "")>
+                            Only share your profile with users who are playing in-game.
+                        </option>
```
Slendy, line 34: Since these are all the same except for the value or selected tag, you could try and simplify this by iterating over all possible values of PrivacyType (or keeping a static cached reference of all types) and then for each option the value would just be the ToString() of the type and the selected would be applied if the users profile visibility equals that type

ProjectLighthouse.Servers.Website/Pages/UserPrivacyPage.cshtml
```diff
@@ -0,0 +1,107 @@
...
+                    </select>
+                </div>
+                <div class="field">
+                    <label style="text-align: left" for="profileCommentsEnabled">
+                        Comments
+                        @if (Model.CommentsDisabledByModerator)
+                        {
+                            <small style="color: red">Locked by a moderator</small>
```
Slendy, line 42: Does Fomantic have classes for formatting text so we don't have to use inline styles?

ProjectLighthouse.Servers.Website/Pages/UserPrivacyPage.cshtml.cs
```diff
@@ -0,0 +1,89 @@
...
+
+    public UserPrivacyPage(DatabaseContext database) : base(database)
+    { }
+
+    public async Task<IActionResult> OnGet([FromRoute] int userId)
+    {
+        this.ProfileUser = await this.Database.Users.FirstOrDefaultAsync(u => u.UserId == userId);
+        if (this.ProfileUser == null) return this.NotFound();
+
+        if (this.User == null) return this.Redirect("~/user/" + userId);
```
Slendy, line 27: this should probably redirect to /login instead

ProjectLighthouse.Servers.Website/Pages/UserPrivacyPage.cshtml.cs
```diff
@@ -0,0 +1,89 @@
...
+        if (this.User != this.ProfileUser) return this.Redirect("~/user/" + userId);
+
+        this.BlockedUsers = await this.Database.BlockedProfiles.Where(b => b.UserId == this.ProfileUser.UserId)
```
Slendy, line 30: I haven't tested this myself yet but does this work without using `.Include(b => b.BlockedUser)` before the select?

ProjectLighthouse.Servers.Website/Pages/UserPrivacyPage.cshtml.cs
```diff
@@ -0,0 +1,89 @@
...
+
+        this.CommentsDisabledByModerator = await this.Database.Cases.Where(c => c.AffectedId == this.ProfileUser.UserId)
+            .Where(c => c.Type == CaseType.UserDisableComments)
+            .Where(c => c.DismissedAt == null)
+            .AnyAsync();
+
+        return this.Page();
+    }
+
+    public async Task<IActionResult> OnPost([FromRoute] int userId, [FromForm] string profilePrivacyLevel, [FromForm] string profileCommentsEnabled, [FromForm] string slotPrivacyLevel)
```
Slendy, line 42: since profileCommentsEnabled is being parsed into true or false anyways does making it a bool work so then ASP.NET will try to implicitly cast it

ProjectLighthouse.Servers.Website/Pages/UserPrivacyPage.cshtml.cs
```diff
@@ -0,0 +1,89 @@
...
+            _ => this.ProfileUser.ProfileVisibility,
+        };
+
+        this.ProfileUser.LevelVisibility = slotPrivacyLevel switch
+        {
+            "public" => PrivacyType.All,
+            "signedInOnly" => PrivacyType.PSN,
+            "inGameOnly" => PrivacyType.Game,
+            _ => this.ProfileUser.LevelVisibility,
+        };
```
Slendy, line 83: Since you use the same function twice you should extract it at least into a local function or a private instance function

ProjectLighthouse.Servers.Website/Pages/UsersPage.cshtml.cs
```diff
@@ -37,8 +37,9 @@ public async Task<IActionResult> OnGet([FromRoute] int pageNumber, [FromQuery] s
...
+            .Where(u => u.Username.Contains(this.SearchValue))
```
Slendy, line 41: The same thing about the chaining style, we should either reformat the entire project or keep using the old style. We shouldn't be mixing

ProjectLighthouse.Servers.Website/Pages/UserPrivacyPage.cshtml.cs
```diff
@@ -0,0 +1,89 @@
...
+                "true" => true,
+                "false" => false,
+                _ => this.ProfileUser.CommentsEnabled,
+            };
+        }
+        else
+        {
+            this.ProfileUser.CommentsEnabled = false;
+        }
+
```
Slendy, line 68: Why is this necessary if users are already prevented from changing whether their comments are enabled

ProjectLighthouse.Servers.Website/Pages/UserPrivacyPage.cshtml.cs
```diff
@@ -0,0 +1,89 @@
...
+        return this.Page();
+    }
+
+    public async Task<IActionResult> OnPost([FromRoute] int userId, [FromForm] string profilePrivacyLevel, [FromForm] string profileCommentsEnabled, [FromForm] string slotPrivacyLevel)
+    {
+        this.ProfileUser = await this.Database.Users.FirstOrDefaultAsync(u => u.UserId == userId);
+        if (this.ProfileUser == null) return this.NotFound();
+
+        if (this.User == null) return this.Redirect("~/user/" + userId);
```
Slendy, line 47: Same thing here, should probably be redirected to /login

ProjectLighthouse.Servers.Website/Pages/UserPage.cshtml.cs
```diff
@@ -128,7 +170,13 @@ public async Task<IActionResult> OnGet([FromRoute] int userId)
             .AnyAsync();
 
         this.IsProfileUserBlocked = await this.Database.IsUserBlockedBy(this.ProfileUser.UserId, this.User.UserId);
-        
+
+        this.CommentsDisabledByModerator = await this.Database.Cases
+            .Where(c => c.AffectedId == this.ProfileUser.UserId)
+            .Where(c => c.Type == CaseType.UserDisableComments)
```
Slendy, line 176: linq style

ProjectLighthouse.Servers.Website/Pages/UserPage.cshtml.cs
```diff
@@ -18,44 +19,81 @@ public class UserPage : BaseLayout
 
     public bool CommentsEnabled;
 
-    public bool IsProfileUserHearted;
+    public List<SlotEntity>? HeartedSlots;
```
Slendy, line 22: Is this meant to be in alphabetical order? If it is, then you should have the same types grouped together then the groups should be sorted alphabetically.

ProjectLighthouse.Servers.Website/Pages/SlotPage.cshtml.cs
```diff
@@ -37,22 +39,28 @@ public async Task<IActionResult> OnGet([FromRoute] int id)
         // Determine if user can view slot according to creator's privacy settings
         if (this.User == null || !this.User.IsAdmin)
         {
-            switch (slot.Creator.ProfileVisibility)
+            switch (slot.Creator.LevelVisibility)
```
Slendy, line 42: This entire calculation of whether a profile should be private or not could be refactored into its own function that takes a privacy level and a user and outputs a bool

Slendy (changes requested):
Everything seems functional, most concerns are clarifications, about styling, or refactoring repeated code

ProjectLighthouse.Servers.Website/Pages/UserPrivacyPage.cshtml.cs
sudokoko: Yes

ProjectLighthouse.Servers.Website/Pages/UserPage.cshtml.cs
```diff
@@ -29,35 +31,25 @@ public class UserPage : BaseLayout
...
-                case PrivacyType.All: break;
-                default: throw new ArgumentOutOfRangeException();
-            }
-        }
+        this.CanAccessProfile = this.ProfileUser.ProfileVisibility.CanAccess(
+            this.User != null, 
+            this.ProfileUser == this.User || this.User != null && this.User.IsModerator);
+        this.CanAccessSlots = this.ProfileUser.LevelVisibility.CanAccess(
+            this.User != null, 
+            this.ProfileUser == this.User || this.User != null && this.User.IsModerator);
```
Slendy, line 52: You could extract the authenticated and owner bools so you don't have to calculate them twice. Something like this: <sub>(please don't commit this directly I'm writing this in a browser)</sub>
```suggestion
        bool isLoggedIn = this.User != null;
        bool isOwner = this.ProfileUser == this.User || this.User != null && this.User.IsModerator;
        this.CanAccessProfile = this.ProfileUser.ProfileVisibility.CanAccess(isLoggedIn, isOwner);
        this.CanAccessSlots = this.ProfileUser.LevelVisibility.CanAccess(isLoggedIn, isOwner);
```

ProjectLighthouse.Servers.Website/Pages/UserPage.cshtml.cs
```diff
@@ -29,35 +31,25 @@ public class UserPage : BaseLayout
     public List<SlotEntity>? QueuedSlots;
 
     public UserEntity? ProfileUser;
+
+    public bool CanAccessProfile;
+    public bool CanAccessSlots;
```
Slendy, line 36: This might infer context better if it were named CanViewProfile and CanViewSlots

ProjectLighthouse.Servers.Website/Pages/SlotPage.cshtml.cs
```diff
@@ -22,6 +22,8 @@ public class SlotPage : BaseLayout
     public bool CommentsEnabled;
     public readonly bool ReviewsEnabled = ServerConfiguration.Instance.UserGeneratedContentLimits.LevelReviewsEnabled;
 
+    public bool CanAccessSlot;
```
Slendy, line 25: Same suggestion, should this be named CanViewSlot?

Slendy (changes requested):
just some nitpicks about naming and duplicate code

Slendy (approved):

ProjectLighthouse/Types/Users/PrivacyType.cs
```diff
@@ -21,6 +21,17 @@ public enum PrivacyType
 
 public static class PrivacyTypeExtensions
 {
+    public static string ToReadableString(this PrivacyType type, string area)
```
Zaprit, line 24: Should this be translated

Zaprit (changes requested):
See comment

ProjectLighthouse/Types/Users/PrivacyType.cs
sudokoko: I have no problem adding translations for it. One moment

Zaprit (approved):
LGTM

PR #846, opened 2023-07-20
Add workaround for the clickable user links in a grief report

sudokoko:
Pretty self explanatory - since the grief report only contains the username of the user instead of the ID, I made the clickable links redirect to the search page for that username.


Slendy (approved):

Zaprit (approved):
LGTM

PR #879, opened 2023-08-23
Add UseApplicationAssets option to RPC API and configuration

### LINKED ISSUE #878: Rich Presence API should explicitly define whether or not assets are Discord assets vs. server assets
**Describe the problem**
Proposed API:
image

We can add a new property to `assets` named `useApplicationAssets` to tell the remote application (e.g. PLRPC) if the hashes below point to Discord hashes added on the application's developer page, or server assets through `/gameAssets/{hash}`

I think Discord assets are far easier for server owners to handle since Discord already provides a nice UI for adding/viewing them, so this solution of defining the behavior server-side will provide compatibility for both.

Both behaviors can technically be calculated client-side by checking what the asset type exactly is, e.g. if it doesn't match the length of a SHA1 hash, then use discord assets, but it's probably better to just explicitly define this behavior server-side.

**API Version(s)**
- /api/v1

**Affected endpoint(s)**
- /rpc

sudokoko:
Closes #878

### DIFF
--- a/ProjectLighthouse.Servers.API/Responses/RpcResponse.cs
+++ b/ProjectLighthouse.Servers.API/Responses/RpcResponse.cs
@@ -19,6 +19,7 @@ public class RpcResponse
             UsernameType = configuration.UsernameType,
             Assets = new RpcAssets
             {
+                UseApplicationAssets = configuration.Assets.UseApplicationAssets,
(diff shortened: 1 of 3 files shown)


Zaprit (approved):
LGTM

PR #880, opened 2023-08-27
Simplify rank badges and implement profile tags

sudokoko:
This PR simplifies the website rank badges to use existing classes, and also implements a small feature that allows server administrators to grant vanity profile tags to users, i.e. for winning an event.

Screenshot 2023-08-26 at 9 35 14 PM

### DIFF
--- a/ProjectLighthouse.Servers.Website/Pages/UserSettingsPage.cshtml.cs
+++ b/ProjectLighthouse.Servers.Website/Pages/UserSettingsPage.cshtml.cs
@@ -20,7 +20,17 @@ public class UserSettingsPage : BaseLayout
     {}
 
     [SuppressMessage("ReSharper", "SpecifyStringComparison")]
-    public async Task<IActionResult> OnPost([FromRoute] int userId, [FromForm] string? avatar, [FromForm] string? username, [FromForm] string? email, [FromForm] string profileTag, [FromForm] string? biography, [FromForm] string? timeZone, [FromForm] string? language)
+    public async Task<IActionResult> OnPost
+    (
+        [FromRoute] int userId,
+        [FromForm] string? avatar,
+        [FromForm] string? username,
+        [FromForm] string? email,
+        [FromForm] string profileTag,
+        [FromForm] string? biography,
+        [FromForm] string? timeZone,
+        [FromForm] string? language
+    )
     {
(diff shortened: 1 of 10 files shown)


Zaprit (approved):
LGTM

ProjectLighthouse.Servers.Website/Pages/UserSettingsPage.cshtml.cs
```diff
@@ -20,7 +20,7 @@ public UserSettingsPage(DatabaseContext database) : base(database)
     {}
 
     [SuppressMessage("ReSharper", "SpecifyStringComparison")]
-    public async Task<IActionResult> OnPost([FromRoute] int userId, [FromForm] string? avatar, [FromForm] string? username, [FromForm] string? email, [FromForm] string? biography, [FromForm] string? timeZone, [FromForm] string? language)
+    public async Task<IActionResult> OnPost([FromRoute] int userId, [FromForm] string? avatar, [FromForm] string? username, [FromForm] string? email, [FromForm] string profileTag, [FromForm] string? biography, [FromForm] string? timeZone, [FromForm] string? language)
```
Slendy, line 23: These parameters should all probably be on separate lines. They were pretty long before but with another field it should probably be wrapped.

ProjectLighthouse.Servers.Website/Pages/UserSettingsPage.cshtml.cs
```diff
@@ -33,6 +33,8 @@ public async Task<IActionResult> OnPost([FromRoute] int userId, [FromForm] strin
 
         if (avatarHash != null) this.ProfileUser.IconHash = avatarHash;
 
+        if (this.User.IsAdmin) this.ProfileUser.ProfileTag = profileTag;
```
Slendy, line 36: Does this also need to check if profile tag isn't null?

Slendy (changes requested):
Mostly LGTM just some small concerns

ProjectLighthouse.Servers.Website/Pages/UserSettingsPage.cshtml.cs
sudokoko: No - if it's null or whitespace it just isn't rendered

Slendy (approved):

PR #891, opened 2023-09-10
Implement dismiss expired cases repeating task

sudokoko:
Self explanatory PR

### DIFF
--- a/ProjectLighthouse.Tests/Unit/ModerationTests.cs
+++ b/ProjectLighthouse.Tests/Unit/ModerationTests.cs
@@ -30,7 +30,7 @@ public class ModerationTests
         DismissExpiredCasesTask task = new();
         await task.Run(database);
         
-        Assert.False(await database.Cases.AnyAsync(c => c.CaseId == 1));
+        Assert.True(await database.Cases.AnyAsync(c => c.CaseId == 1 && c.DismissedAt != null));
     }
 
     [Fact]
@@ -52,6 +52,6 @@ public class ModerationTests
         DismissExpiredCasesTask task = new();
         await task.Run(database);
 
-        Assert.True(await database.Cases.AnyAsync(c => c.CaseId == 2));
+        Assert.True(await database.Cases.AnyAsync(c => c.CaseId == 2 && c.DismissedAt == null));
     }
 }
--- a/ProjectLighthouse.Tests/Unit/ModerationTests.cs
+++ b/ProjectLighthouse.Tests/Unit/ModerationTests.cs
@@ -29,8 +29,8 @@ public class ModerationTests
         
         DismissExpiredCasesTask task = new();
         await task.Run(database);
-        
-        Assert.True(await database.Cases.AnyAsync(c => c.CaseId == 1 && c.DismissedAt != null));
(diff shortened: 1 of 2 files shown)


sudokoko:
zaprit wants me to write a unit test

ProjectLighthouse.Tests/Unit/ModerationTests.cs
```diff
@@ -0,0 +1,57 @@
...
+{
+    [Fact]
+    public async void CanDismissExpiredCases()
+    {
+        await using DatabaseContext database = await MockHelper.GetTestDatabase();
+        
+        ModerationCaseEntity @case = new()
+        {
+            CaseId = 1,
+            ExpiresAt = DateTime.Now - TimeSpan.FromHours(1),
```
Zaprit, line 22: This could also just use the epoch

ProjectLighthouse/Administration/Maintenance/RepeatingTasks/DismissExpiredCasesTask.cs
```diff
@@ -0,0 +1,39 @@
...
+
+public class DismissExpiredCasesTask : IRepeatingTask
+{
+    public string Name => "Dismiss Expired Cases";
+    public TimeSpan RepeatInterval => TimeSpan.FromHours(4);
+    public DateTime LastRan { get; set; }
+
+    public async Task Run(DatabaseContext database)
+    {
+        List<ModerationCaseEntity> expiredCases =
```
Zaprit, line 22: Does this account for timezones?

Zaprit: it may not be an issue as long as the time in the case is set by the server, though DST may still be a problem

Zaprit (changes requested):
See comments

ProjectLighthouse/Administration/Maintenance/RepeatingTasks/DismissExpiredCasesTask.cs
sudokoko: Is it not stored in UTC?

ProjectLighthouse.Tests/Unit/ModerationTests.cs
sudokoko: For what reason

ProjectLighthouse/Administration/Maintenance/RepeatingTasks/DismissExpiredCasesTask.cs
sudokoko: And even then it's not been an issue in any other aspect of the mod system such as case expiration so I would assume it'd be fine

ProjectLighthouse.Tests/Unit/ModerationTests.cs
Zaprit: Less math.

Slendy (approved):
lgtm

Zaprit (approved):
LGTM

PR #900, opened 2023-09-13
Improve moderation case page and case partial

sudokoko:
This PR adds a few quality of life features and fixes to the moderation case page and case partial.

* Adds pagination to the moderation case page to reduce database strain for instances with large amounts of cases
* Fully implements the search functionality of the moderation case page
* Extends and simplifies the case status indicators, basically a tail end to #891
* Disable the ability to dismiss cases "early" after they've been expired to allow for a seamless process of dismissal
* Turned down the repeat time of the dismissal task to every 1 hour

Technically outside the scope of this PR, but also throws in pagination and search functionality into the admin users list.

### DIFF
--- a/ProjectLighthouse.Servers.Website/Pages/Moderation/CasePage.cshtml.cs
+++ b/ProjectLighthouse.Servers.Website/Pages/Moderation/CasePage.cshtml.cs
@@ -50,7 +50,7 @@ public class CasePage : BaseLayout
 
         this.Cases = await this.Database.Cases.Include(c => c.Creator)
             .Include(c => c.Dismisser)
-            .Where(c => c.Reason.Contains(this.SearchValue))
+            .Where(c => c.AffectedId.ToString().Contains(this.SearchValue))
             .OrderByDescending(c => c.CaseId)
             .Skip(pageNumber * ServerStatics.PageSize)
             .Take(ServerStatics.PageSize)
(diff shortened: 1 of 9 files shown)


Zaprit (approved):
LGTM

ProjectLighthouse.Servers.Website/Pages/Moderation/CasePage.cshtml.cs
```diff
@@ -31,18 +32,27 @@ public async Task<IActionResult> OnGet([FromRoute] int pageNumber, [FromQuery] s
...
-            .ToListAsync();
-        
-        this.CaseCount = await this.Database.Cases.CountAsync(c => c.Reason.Contains(this.SearchValue));
-        this.DismissedCaseCount = await this.Database.Cases.CountAsync(c => c.Reason.Contains(this.SearchValue) && c.DismissedAt != null);
+        this.CaseCount =
+            await this.Database.Cases.CountAsync(c => c.Reason.Contains(this.SearchValue));
+        this.ExpiredCaseCount =
+            await this.Database.Cases.CountAsync(c => c.Reason.Contains(this.SearchValue) && c.DismissedAt == null && c.ExpiresAt < DateTime.UtcNow);
+        this.DismissedCaseCount =
+            await this.Database.Cases.CountAsync(c => c.Reason.Contains(this.SearchValue) && c.DismissedAt != null);
```
Slendy, line 40: Why does this use the case reason rather than like the person who the case is against? Or this could potentially search against multiple fields like the slot search.

Slendy (changes requested):

ProjectLighthouse.Servers.Website/Pages/Moderation/CasePage.cshtml.cs
sudokoko: Typo, most likely. I'll correct it.

Slendy (approved):

PR #906, opened 2023-09-22
Move API authentication logic into DatabaseContext partial class

sudokoko:
This should allow for easier authentication of API calls in the future when it becomes needed.


Slendy (approved):

Zaprit (approved):
Look good me code very nice should merge 👍

PR #909, opened 2023-09-30
Add links to license and source code in footer

sudokoko:
Adds hyperlinks to the upstream LICENSE file, as well as the active git remote, to help better comply with AGPL's requirements.

Screenshot 2023-09-30 at 10 57 52 AM

### DIFF
--- a/ProjectLighthouse.Servers.Website/Pages/Layouts/BaseLayout.cshtml
+++ b/ProjectLighthouse.Servers.Website/Pages/Layouts/BaseLayout.cshtml
@@ -199,7 +199,7 @@
                     <p>@Model.Translate(BaseLayoutStrings.GeneratedModified)</p>
                 }
                 <span style="float: right;">
-                    <a href="https://github.com/LBPUnion/ProjectLighthouse/blob/main/LICENSE">
+                    <a href="https://github.com/LBPUnion/ProjectLighthouse/blob/main/LICENSE" target="_blank">
                         AGPL-3.0 License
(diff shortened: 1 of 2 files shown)


ProjectLighthouse.Servers.Website/Pages/Layouts/BaseLayout.cshtml
```diff
@@ -191,12 +191,22 @@
...
                 @if (VersionHelper.IsDirty)
                 {
                     <p>@Model.Translate(BaseLayoutStrings.GeneratedModified)</p>
                 }
+                <span style="float: right;">
+                    <a href="https://github.com/LBPUnion/ProjectLighthouse/blob/main/LICENSE" target="_blank">
+                        AGPL-3.0 License
+                    </a>
+                    &centerdot;
+                    <a href="@VersionHelper.Remotes[0].Split(".git")[0]" target="_blank">
```
Slendy, line 206: Is there a chance this could throw an error if it were built improperly? I think this could be cleaned up a little to use more linq instead of array indexes. Also, have you checked that this functions properly when there is more than 1 remote?

Slendy (changes requested):

sudokoko:
In general the version helper could probably be replaced with something like [GitInfo](https://github.com/devlooped/GitInfo) in the future but otherwise this works

Slendy (approved):

Zaprit (approved):
LGTM

PR #911, opened 2023-10-04
Migrate to GitInfo for embedding git information

Slendy:
Gets rid of the old system that would copy .txt files to the output directory. Also gets rid of the RazorRuntimeCompilation because apparently it isn't needed for hot reload and was only used for development anyways


ProjectLighthouse/Helpers/VersionHelper.cs
```diff
@@ -12,64 +11,28 @@ static VersionHelper()
     {
         try
         {
-            CommitHash = ResourceHelper.ReadManifestFile("gitVersion.txt");
-            Branch = ResourceHelper.ReadManifestFile("gitBranch.txt");
+            CommitHash = ThisAssembly.Git.Commit;
+            Branch = ThisAssembly.Git.Branch;
             string commitNumber = $"{CommitHash}_{Build}";
             FullRevision = Branch == "main" ? $"r{commitNumber}" : $"{Branch}_r{commitNumber}";
```
sudokoko, line 17: Is the try/catch still necessary here seeing as these values are constants?

sudokoko (changes requested):

ProjectLighthouse/Helpers/VersionHelper.cs
TorutheRedFox: > Is the try/catch still necessary here seeing as these values are constants?

yeah it seems the try/catch was entirely there to handle errors caused by missing/malformed txt files

Slendy: Removed the try catch and the entire constructor since it was unnecessary

sudokoko (approved):

Zaprit (approved):
LGTM

PR #930, opened 2023-10-24
Add checks for resource strings

Slendy:
Pretty self-explanable. This PR prevents people from trying to put payloads in resource hashes like path traversals or HTML escapes. I don't think this needs an entire migration, it should be relatively simple for server owners to fix any occurrences of invalid resources and I don't think the issue is that widespread at the moment.


Zaprit (approved):
LGTM

sudokoko (approved):
this looks fine

PR #935, opened 2023-10-29
Missing digest message change

Zaprit:
This PR changes the message that shows when the digestKey has not been set to better reflect what will/won't work if left unset.

### DIFF
--- a/ProjectLighthouse.Servers.GameServer/Startup/GameServerStartup.cs
+++ b/ProjectLighthouse.Servers.GameServer/Startup/GameServerStartup.cs
@@ -90,7 +90,7 @@ public class GameServerStartup
         {
             Logger.Warn
             (
-                "The serverDigestKey configuration option wasn't set, so digest headers won't be set or verified. This will prevent LBP 1, and LBP 3 from working, and may break features in the other games (i.e Dive in). " +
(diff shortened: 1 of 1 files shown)


ProjectLighthouse.Servers.GameServer/Startup/GameServerStartup.cs
```diff
@@ -90,7 +90,7 @@ public virtual void Configure(IApplicationBuilder app, IWebHostEnvironment env)
         {
             Logger.Warn
             (
-                "The serverDigestKey configuration option wasn't set, so digest headers won't be set or verified. This will also prevent LBP 1, LBP 2, and LBP Vita from working. " +
+                "The serverDigestKey configuration option wasn't set, so digest headers won't be set or verified. This will prevent LBP 1, and LBP 3 from working, and may break features in the other games (i.e Dive in). " +
```
sudokoko, line 93:
```suggestion
                "The serverDigestKey configuration option wasn't set, so digest headers won't be set or verified. This will prevent LBP 1 and LBP 3 from working, and may break features in the other games (i.e Dive In). " +
```

sudokoko (approved):
minor grammatical nitpick otherwise LGTM

Slendy (approved):

PR #932, opened 2023-10-26
Implement in-game and website notifications

sudokoko:
This PR implements the `/notification` endpoint, as well as merges the website announcements page and a new section dedicated for notifications together.

Also implements the ability to return manually created XML as a valid response. This is used here to stack notifications.

Obviously, not all notification types are implemented here. This implements notifications for user/slot moderation, team picks, and level publish issues/failures.

The `SendNotification(int userId, string text, NotificationType type)` method within the `DatabaseContext` class can be used to easily queue notifications.

- [x] Implement in-game notifications
- [x] Implement ability to return manually created XML as valid response
- [x] Implement website notifications
  - [x] Merge announcements/notifications together

PR #937, opened 2023-10-29
Add missing moderation notification calls

sudokoko:
Extends #932. Adds notifications for photo, score, comment, and review deletions.

### DIFF
--- a/ProjectLighthouse.Servers.Website/Controllers/Moderator/ModerationRemovalController.cs
+++ b/ProjectLighthouse.Servers.Website/Controllers/Moderator/ModerationRemovalController.cs
@@ -56,8 +56,9 @@ public class ModerationRemovalController : ControllerBase
         UserEntity? user = this.database.UserFromWebRequest(this.Request);
         if (user == null) return this.Redirect("~/login");
 
-        CommentEntity? comment = await this.database.Comments.Include(commentEntity => commentEntity.TargetUser)
-            .Include(commentEntity => commentEntity.TargetSlot)
+        CommentEntity? comment = await this.database.Comments
+            .Include(c => c.TargetUser)
+            .Include(c => c.TargetSlot)
             .FirstOrDefaultAsync(c => c.CommentId == commentId);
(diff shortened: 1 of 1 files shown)


ProjectLighthouse.Servers.Website/Controllers/Moderator/ModerationRemovalController.cs
```diff
@@ -50,7 +56,10 @@ public async Task<IActionResult> DeleteComment(int commentId, [FromQuery] string
         UserEntity? user = this.database.UserFromWebRequest(this.Request);
         if (user == null) return this.Redirect("~/login");
 
-        CommentEntity? comment = await this.database.Comments.FirstOrDefaultAsync(c => c.CommentId == commentId);
+        CommentEntity? comment = await this.database.Comments
+            .Include(c => c.TargetUser)
+            .Include(c => c.TargetSlot)
```
Slendy, line 61: Is this confirmed to be working without issue, I don't know how EF handles including a nullable foreign key

Slendy (approved):

ProjectLighthouse.Servers.Website/Controllers/Moderator/ModerationRemovalController.cs
sudokoko: To my knowledge including a nullable foreign key works fine? I'll double check though

sudokoko: I don't see why this wouldn't work, see below

https://github.com/LBPUnion/ProjectLighthouse/blob/571986c384d04903b89ee240c67e1be05b7e8a31/ProjectLighthouse.Servers.Website/Pages/NotificationsPage.cshtml.cs#L24-L25

https://github.com/LBPUnion/ProjectLighthouse/blob/571986c384d04903b89ee240c67e1be05b7e8a31/ProjectLighthouse/Types/Entities/Website/WebsiteAnnouncementEntity.cs#L20-L21

PR #940, opened 2023-10-30
Miscellaneous website/moderation/notification fixes

sudokoko:
This PR addresses a few minor issues and fixes a few things.

* Sub-levels would be displayed on a user's profile when logged out
* Fixes an issue when trying to directly access a sub-level page on the website as the slot creator
* Mark automatically dismissed cases as unprocessed so the repeating task will perform case actions
* Use a foreach for notifications
  * It was thought that modifying a collection item while iterating over the list would throw an exception but this is exactly what the automatic case dismissal repeating task does. Tested this and it works perfectly fine.


W0lf4llo:
LGTM 👍

Zaprit (approved):
LGTM

PR #965, opened 2024-01-15
Use UtcNow instead of Now for internal time keeping

Slendy:
`DateTime.Now` uses local machine time which would potentially cause discrepancies with timekeeping for token or case expirations.


sudokoko (approved):
Is there any way we can migrate existing times to use UtcNow? Or is that not an issue? Otherwise this looks fine.

Zaprit:
will this affect any level upload times or no?

Slendy:
> will this affect any level upload times or no?

No, those were always stored as UTC since they used the `TimeHelper` functions.

Zaprit (approved):

PR #966, opened 2024-01-15
Migrate to SQLite EF provider for testing

Slendy:
Currently, we use the InMemory EF provider which has some issues like not enforcing referential integrity (a level can exist without a creator). It also doesn't detect potential issues with database schema. Switching to in-memory SQLite fixes both these issues and will make unit tests more realistic.

### DIFF
--- a/ProjectLighthouse.Tests.GameApiTests/Unit/Controllers/CommentControllerTests.cs
+++ b/ProjectLighthouse.Tests.GameApiTests/Unit/Controllers/CommentControllerTests.cs
@@ -94,7 +94,7 @@ public class CommentControllerTests
                 Type = SlotType.User,
             },
         };
-        await using DatabaseContext dbMock = await MockHelper.GetTestDatabase(new[]{slots,});
+        await using DatabaseContext dbMock = await MockHelper.GetTestDatabase(slots);
 
         CommentController commentController = new(dbMock);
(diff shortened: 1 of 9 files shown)


Zaprit:
Why is CI fail????? 🧄

sudokoko:
> Why is CI fail????? 🧄 On a more serious note, how hard would it be to also add... for instance PostgreSQL support while we're at it?

This would require multiple different migrations for the same change no?

Slendy:
> Why is CI fail????? 🧄 On a more serious note, how hard would it be to also add... for instance PostgreSQL support while we're at it?

If you were to change the codebase to use Postgres instead of MySQL it would be relatively simple because of EntityFramework. You would just have to delete all the previous MySQL migrations and create a new initial migration for Postgres. The hard part is migrating an existing MySQL Lighthouse instance to Postgres. If you keep your instance up to date it would only require some manual migration to transfer your data from MySQL into a Postgres database which could be done with a script (one that we might have to write since there are some incompatible types). If you didn't keep your instance up to date then you would have to instruct people to update to the latest commit before the database switch to apply all the missing migrations and then they could migrate using the previously stated method. I would only consider switching if there were some solid justifications for doing so.

W0lf4llo (commented):
LGTM!

sudokoko (approved):
This looks fine.

Zaprit (approved):
LGTM!

PR #970, opened 2024-01-20
Move StaticFiles from the base project to the Website project

Slendy:
# This could be a breaking change for some installations

If you rely on a symbolic link or a manual copy operation to put website files in the data directory you will need to change your installation to instead copy from `ProjectLighthouse.Servers.Website\StaticFiles`. I'm still not sure if this is the best approach for storing these static files. I think this could be the opportunity to include them in the output of the build and to pull from that instead of them being in the data directory. I'm open to hearing discussion from others though.


sudokoko (approved):
feel free to re-request review if you end up changing something

Zaprit (approved):

PR #980, opened 2024-02-14
Sort team picks by the time they were team picked

Slendy:
Instead of sorting team picks by the time the level was last updated or uploaded, we can store when the level was picked and sort by that instead. On migration, already team-picked levels are set to a timestamp of 1 so they will still count as team-picked but will all show up last.


sudokoko (approved):

Zaprit (approved):
LGTM, I would personally have left the bool in as it probably indexes better but ehh, should be fine.

CrankyLandlord:
yaaaaay. The creators thank you all 👍👍

PR #990, opened 2024-03-01
Rewrite digest middleware

Slendy:
This PR is a refactor of the digest middleware. The new version makes it much less dependent on hardcoded endpoints and the digests in the config. Now the hardcoded behavior can be controlled by changing properties in the `UseDigestAttribute` class, like whether or not the request should fail if the client sends an incorrect digest or what the digest header name should be. The middleware can now also use unlimited digest keys to check against. 

This PR also adds a new GameController class which extracts most of the common attributes the LBP endpoints use so they don't have to be redefined in every file.

### DIFF
--- a/ProjectLighthouse.Servers.GameServer/Startup/GameServerStartup.cs
+++ b/ProjectLighthouse.Servers.GameServer/Startup/GameServerStartup.cs
@@ -84,8 +84,6 @@ public class GameServerStartup
     // This method gets called by the runtime. Use this method to configure the HTTP request pipeline.
     public virtual void Configure(IApplicationBuilder app, IWebHostEnvironment env)
     {
-        bool computeDigests = true;
-
         if (string.IsNullOrEmpty(ServerConfiguration.Instance.DigestKey.PrimaryDigestKey))
         {
             Logger.Warn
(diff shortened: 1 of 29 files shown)


ProjectLighthouse.Servers.GameServer/Startup/GameServerStartup.cs
```diff
@@ -105,11 +102,17 @@ public virtual void Configure(IApplicationBuilder app, IWebHostEnvironment env)
...
         app.UseMiddleware<SetLastContactMiddleware>();
 
         app.UseRouting();
 
+        List<string> digestKeys =
+        [
+            ServerConfiguration.Instance.DigestKey.PrimaryDigestKey,
+            ServerConfiguration.Instance.DigestKey.AlternateDigestKey,
+        ];
+        app.UseMiddleware<DigestMiddleware>(digestKeys);
```
sudokoko, line 114: is there a reason to moving this below UseRouting that i'm missing?

Slendy: It's so that the middleware can access the endpoint data like the attributes. If you put it before the UseRouting the endpoint metadata isn't loaded yet.

sudokoko:
merge conflicts and @Zaprit review bump

PR #855, opened 2023-08-04
Implement in-game recent activity

Slendy:
Website recent activity still needs to be done, ideally, once #839 has been merged.

## Todos
- [x] Cleanup code and remove debug logging
- [x] Test at a larger scale
- [x] Write more unit tests
- [x] Add timestamp to website announcements
- [x] ~~Don't track unhearts cause the game doesn't actually parse them~~ (going to leave this in for website activity)
- [x] Merge #838 and #830

Closes #5, closes #8

PR #1001, opened 2024-03-28
Implement read-only mode

sudokoko:
This PR implements the ability to put an instance in a "read-only" state.

The following actions are limited or disabled while in read-only mode:
* Level uploads or changes
* Photo uploads
* General asset uploads (/upload/hash)
* Profile picture/biography changes
* Post/delete comments
* Post/delete reviews

A message will also be appended to the base website layout, as well as the /announce endpoint.

This could probably be refactored into a simple middleware or something in the future.

### DIFF
--- a/ProjectLighthouse.Servers.Website/Controllers/UserPageController.cs
+++ b/ProjectLighthouse.Servers.Website/Controllers/UserPageController.cs
@@ -40,7 +40,7 @@ public class UserPageController : ControllerBase
         if (token == null) return this.Redirect("~/login");
 
         // Deny request if in read-only mode
-        if (ServerConfiguration.Instance.UserGeneratedContentLimits.ReadOnlyMode) return this.BadRequest();
+        if (ServerConfiguration.Instance.UserGeneratedContentLimits.ReadOnlyMode) return this.Redirect("~/user/" + id);
 
         if (msg == null)
         {
--- a/ProjectLighthouse.Servers.Website/Controllers/UserPageController.cs
+++ b/ProjectLighthouse.Servers.Website/Controllers/UserPageController.cs
@@ -39,6 +39,9 @@ public class UserPageController : ControllerBase
         WebTokenEntity? token = this.database.WebTokenFromRequest(this.Request);
         if (token == null) return this.Redirect("~/login");
 
+        // Deny request if in read-only mode
+        if (ServerConfiguration.Instance.UserGeneratedContentLimits.ReadOnlyMode) return this.BadRequest();
+
         if (msg == null)
         {
             Logger.Error($"Refusing to post comment from {token.UserId} on user {id}, {nameof(msg)} is null", LogArea.Comments);
--- a/ProjectLighthouse.Servers.Website/Pages/UserSettingsPage.cshtml.cs
+++ b/ProjectLighthouse.Servers.Website/Pages/UserSettingsPage.cshtml.cs
@@ -39,15 +39,13 @@ public class UserSettingsPage : BaseLayout
 
         if (!this.User.IsModerator && this.User != this.ProfileUser) return this.Redirect("~/user/" + userId);
 
+        // Deny request if in read-only mode
+        if (avatar != null && ServerConfiguration.Instance.UserGeneratedContentLimits.ReadOnlyMode)
+            return this.Redirect($"~/user/{userId}");
+
         string? avatarHash = await FileHelper.ParseBase64Image(avatar);
 
-        if (avatarHash != null)
-        {
-            // Deny request if in read-only mode
-            if (ServerConfiguration.Instance.UserGeneratedContentLimits.ReadOnlyMode) return this.Redirect($"~/user/{userId}");
-
-            this.ProfileUser.IconHash = avatarHash;
-        }
+        if (avatarHash != null) this.ProfileUser.IconHash = avatarHash;
 
         if (this.User.IsAdmin) this.ProfileUser.ProfileTag = profileTag;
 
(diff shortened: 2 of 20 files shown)


ProjectLighthouse.Servers.Website/Pages/UserSettingsPage.cshtml.cs
```diff
@@ -41,12 +41,21 @@ public UserSettingsPage(DatabaseContext database) : base(database)
 
         string? avatarHash = await FileHelper.ParseBase64Image(avatar);
 
-        if (avatarHash != null) this.ProfileUser.IconHash = avatarHash;
+        if (avatarHash != null)
+        {
+            // Deny request if in read-only mode
+            if (ServerConfiguration.Instance.UserGeneratedContentLimits.ReadOnlyMode) return this.BadRequest();
```
Slendy, line 47:
```suggestion
            if (ServerConfiguration.Instance.UserGeneratedContentLimits.ReadOnlyMode) return this.Redirect($"~/users/{userId}");
```
This should probably redirect back to the user's page rather than returning a bad request so that they don't get stuck on the 400 page.

ProjectLighthouse.Servers.Website/Pages/UserSettingsPage.cshtml.cs
```diff
@@ -41,12 +41,21 @@ public UserSettingsPage(DatabaseContext database) : base(database)
...
+
+            this.ProfileUser.IconHash = avatarHash;
+        }
 
         if (this.User.IsAdmin) this.ProfileUser.ProfileTag = profileTag;
 
         if (biography != null)
         {
+            // Deny request if in read-only mode
+            if (ServerConfiguration.Instance.UserGeneratedContentLimits.ReadOnlyMode) return this.BadRequest();
```
Slendy, line 57:
```suggestion
            if (ServerConfiguration.Instance.UserGeneratedContentLimits.ReadOnlyMode) return this.Redirect($"~/users/{userId}");
```

ProjectLighthouse.Servers.Website/Pages/UserSettingsPage.cshtml.cs
Slendy, line 47: This check should probably be above `FileHelper.ParseBase64Image` since that method will immediately add an asset if it is valid

Slendy (changes requested):
Some small feedback about returning 400 on the website page. You also need to add checks in `SlotSettingsPage` and protect the calls to `ParseBase64Image`

ProjectLighthouse.Servers.Website/Pages/UserSettingsPage.cshtml.cs
sudokoko: My only concern with this is that I will have to check if the form value `avatar` is not null. Does this value get submitted even if no new avatar is selected?

Slendy: If there's no new avatar then the form value is empty, not sure if that translates to null in the C# controller though. We might be better off just disabling this endpoint in read-only mode unless you're an admin

sudokoko: > not sure if that translates to null in the C# controller though.

It should, it is handled by the `ParseBase64Image` method

https://github.com/LBPUnion/ProjectLighthouse/blob/f1c5ad4002ab0a673c7881dcc9c5d1f44f311ddc/ProjectLighthouse/Files/LbpImageHelper.cs#L34

sudokoko:
> Looks good, however do you think it would be worth while having some additional information field in there, so we can link to the blog post?

I think that might be a bit too specific to our case, though feel free to object.

HenryAsbridge:
> specific

I suspect that anyone who's putting their lighthouse instance into read only mode is doing so for a reason, and may want to communicate with their users as to why

Zaprit:
> > specific
> 
> I suspect that anyone who's putting their lighthouse instance into read only mode is doing so for a reason, and may want to communicate with their users as to why

and this is why you don't use the github mobile app, because it will make you respond on your work account smh. why can't it use the account that's in the org where the PR is, rather than the wrong account at the wrong time

Zaprit:
> > > specific
> > 
> > 
> > I suspect that anyone who's putting their lighthouse instance into read only mode is doing so for a reason, and may want to communicate with their users as to why
> 
> [#1001 (comment)](https://github.com/LBPUnion/ProjectLighthouse/pull/1001#issuecomment-2025029186)
> 
> > An instance administrator can always link to it in the announcement config and create a web based announcement.
> 
> I may just be OCD, but I feel like implementing a dedicated configuration option for something that can be done through existing config options (announceText) and website features (web announcements, which will be implemented into the game as well through #855) is a bit redundant. Though if this is something you wholeheartedly believe should be implemented, I will do so. Just thought I'd voice my concern on it before doing that.

I suppose the important part is do our banner announcements show up front and centre on the home page of the instance in a way that doesn't require you to go digging?

sudokoko:
> I suppose the important part is do our banner announcements show up front and centre on the home page of the instance in a way that doesn't require you to go digging?

No, however I would not be opposed to implementing this, and would actually prefer doing it this way. I would probably just package those changes in with this PR since they are technically in scope.

Zaprit (approved):

ProjectLighthouse.Servers.Website/Controllers/UserPageController.cs
```diff
@@ -39,6 +39,9 @@ public async Task<IActionResult> PostComment([FromRoute] int id, [FromForm] stri
         WebTokenEntity? token = this.database.WebTokenFromRequest(this.Request);
         if (token == null) return this.Redirect("~/login");
 
+        // Deny request if in read-only mode
+        if (ServerConfiguration.Instance.UserGeneratedContentLimits.ReadOnlyMode) return this.BadRequest();
```
Slendy, line 43: This should probably redirect back to the original page instead of returning bad request since it’s SSR.

Slendy (changes requested):
Only issue is returning a 400 status code on the website instead of redirecting back to the original page to stay consistent with the rest of the website.

Slendy (approved):

PR #1015, opened 2024-04-24
Prevent LBP3 reviews from showing up in LBP2

Slendy:
There is currently a bug where if you review an LBP3 level and load your profile in LBP2 the game freaks out because the level data won't be returned because of the GameVersionFilter. This PR adds similar filtering so that reviews from higher game versions are no longer returned. This change adds an extra SQL join to the review query.

new example generated query:
```sql
.param set @__targetUserId_0 1
.param set @__token_GameVersion_1 1

SELECT "r"."ReviewId", "r"."Deleted", "r"."DeletedBy", "r"."LabelCollection", "r"."ReviewerId", "r"."SlotId", "r"."Text", "r"."Thumb", "r"."ThumbsDown", "r"."ThumbsUp", "r"."Timestamp"
FROM "Reviews" AS "r"
INNER JOIN "Slots" AS "s" ON "r"."SlotId" = "s"."SlotId"
WHERE "r"."ReviewerId" = @__targetUserId_0 AND "s"."GameVersion" <= @__token_GameVersion_1
ORDER BY "r"."Timestamp" DESC
```

### DIFF
--- a/ProjectLighthouse.Servers.GameServer/Controllers/Slots/ReviewController.cs
+++ b/ProjectLighthouse.Servers.GameServer/Controllers/Slots/ReviewController.cs
@@ -157,6 +157,13 @@ public class ReviewController : ControllerBase
 
         List<GameReview> reviews = (await this.database.Reviews
             .Where(r => r.SlotId == slotId)
+            .Select(r => new
+            {
+                Review = r,
+                SlotVersion = r.Slot!.GameVersion,
+            })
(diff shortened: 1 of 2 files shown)


sudokoko (approved):
this looks fine

Zaprit (approved):
Idk, kinda looks like code to me
tenor-ezgif com-added-text(1)

PR #1022, opened 2024-05-31
Implement the ability to forcibly verify a user's email

sudokoko:
Self explanatory PR.


Zaprit (changes requested):

Slendy (approved):
Lgtm see zaprit’s comment though. You could also remove all the existing tokens you fetched after verifying the email but I don’t think it makes a difference since they’ll expire eventually.

Zaprit (approved):
Ok

PR #1027, opened 2024-06-14
Implement the ability for moderators to delete all scores/comments by a user

Metraberryy:
Title is fairly self-explanatory.
This PR allows moderators/admins to delete every score or comment posted by a user from the user page.
These features were requested by a Beacon moderator after some recent comment and hacked score spam.

### DIFF
--- a/ProjectLighthouse.Servers.Website/Controllers/Admin/AdminUserController.cs
+++ b/ProjectLighthouse.Servers.Website/Controllers/Admin/AdminUserController.cs
@@ -118,8 +118,6 @@ public class AdminUserController : ControllerBase
         await this.database.SendNotification(targetedUser.UserId,
             "Your comments have been deleted by a moderator.");
 
-        await this.database.SaveChangesAsync();
-
         return this.Redirect($"/user/{targetedUser.UserId}");
     }
 
@@ -140,8 +138,6 @@ public class AdminUserController : ControllerBase
         Logger.Success($"Deleted scores for {targetedUser.Username} (id:{targetedUser.UserId})", LogArea.Admin);
 
         await this.database.SendNotification(targetedUser.UserId, "Your scores have been deleted by a moderator.");
-
-        await this.database.SaveChangesAsync();
 
         return this.Redirect($"/user/{targetedUser.UserId}");
     }
--- a/ProjectLighthouse.Servers.Website/Controllers/Admin/AdminUserController.cs
+++ b/ProjectLighthouse.Servers.Website/Controllers/Admin/AdminUserController.cs
@@ -118,9 +118,8 @@ public class AdminUserController : ControllerBase
             comment.Deleted = true;
             comment.DeletedBy = user.Username;
             comment.DeletedType = "moderator";
-            Logger.Success($"Deleted comments for {targetedUser.Username} (id:{targetedUser.UserId})",
-                LogArea.Admin);
         }
+        Logger.Success($"Deleted comments for {targetedUser.Username} (id:{targetedUser.UserId})", LogArea.Admin);
 
         await this.database.SendNotification(targetedUser.UserId,
             "Your comments have been deleted by a moderator.");
--- a/ProjectLighthouse.Servers.Website/Controllers/Admin/AdminUserController.cs
+++ b/ProjectLighthouse.Servers.Website/Controllers/Admin/AdminUserController.cs
@@ -1,7 +1,9 @@
 #nullable enable
 using LBPUnion.ProjectLighthouse.Database;
+using LBPUnion.ProjectLighthouse.Extensions;
 using LBPUnion.ProjectLighthouse.Files;
 using LBPUnion.ProjectLighthouse.Logging;
+using LBPUnion.ProjectLighthouse.Types.Entities.Level;
 using LBPUnion.ProjectLighthouse.Types.Entities.Profile;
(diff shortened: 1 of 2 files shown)


ProjectLighthouse.Servers.Website/Controllers/Admin/AdminUserController.cs
```diff
@@ -91,6 +93,63 @@ public AdminUserController(DatabaseContext database)
 
         return this.Redirect($"/user/{targetedUser.UserId}");
     }
+    
+    /// <summary>
+    /// Deletes every comment by the user. Useful in case of mass spam
+    /// </summary>
+    [HttpGet("wipeComments")]
+    public async Task<IActionResult> WipeComments([FromRoute] int id) {
```
sudokoko, line 101:
```suggestion
    public async Task<IActionResult> WipeComments([FromRoute] int id)
    {
```

sudokoko (approved):
lgtm, bracket formatting nitpick

ProjectLighthouse.Servers.Website/Controllers/Admin/AdminUserController.cs
```diff
@@ -91,6 +93,63 @@ public AdminUserController(DatabaseContext database)
...
+        UserEntity? user = this.database.UserFromWebRequest(this.Request);
+        if (user == null || !user.IsModerator) return this.NotFound();
+
+        UserEntity? targetedUser = await this.database.Users.FirstOrDefaultAsync(u => u.UserId == id);
+        if (targetedUser == null) return this.NotFound();
+
+        // Get every comment posted by the target user
+        List<CommentEntity> comments = await this.database.Comments
+            .Where(c => c.Poster == targetedUser)
```
Slendy, line 112: This might be some premature optimization but EntityFramework has support for bulk updates which can be done like this
```csharp
await database.Comments.Where(c => c.PosterUserId == targetUser.UserId)
            .ExecuteUpdateAsync(s =>
                   s.SetProperty(c => c.Deleted, true)
                    .SetProperty(c => c.DeletedBy, user.Username)
                    .SetProperty(c => c.DeletedType, "moderator"));
```
I think the for loop is more readable IMO but you would have to performance test to see how each method would scale.

ProjectLighthouse.Servers.Website/Controllers/Admin/AdminUserController.cs
```diff
@@ -91,6 +94,64 @@ public AdminUserController(DatabaseContext database)
...
+            .ToListAsync();
+
+        // Loop through and delete all of the comments
+        foreach (CommentEntity comment in comments)
+        {
+            comment.Deleted = true;
+            comment.DeletedBy = user.Username;
+            comment.DeletedType = "moderator";
+            Logger.Success($"Deleted comments for {targetedUser.Username} (id:{targetedUser.UserId})",
```
Slendy, line 121: This should probably be outside the loop otherwise it could cause log spam for users with a lot of comments.

ProjectLighthouse.Servers.Website/Controllers/Admin/AdminUserController.cs
```diff
@@ -91,6 +94,64 @@ public AdminUserController(DatabaseContext database)
...
+    public async Task<IActionResult> WipeScores([FromRoute] int id)
+    {
+        UserEntity? user = this.database.UserFromWebRequest(this.Request);
+        if (user == null || !user.IsModerator) return this.NotFound();
+
+        UserEntity? targetedUser = await this.database.Users.FirstOrDefaultAsync(u => u.UserId == id);
+        if (targetedUser == null) return this.NotFound();
+
+        // Find and delete every score uploaded by the target user
+        await this.database.Scores.RemoveWhere(c => c.User == targetedUser);
```
Slendy, line 146: For this, you could probably use a batch delete and use the UserId property rather than the foreign key property.
```csharp
await this.database.Scores.Where(c => c.UserId == targetedUser.UserId).ExecuteDeleteAsync();
```

ProjectLighthouse.Servers.Website/Controllers/Admin/AdminUserController.cs
Metraberryy: i dunno, but i like this better anyway

ProjectLighthouse.Servers.Website/Controllers/Admin/AdminUserController.cs
```diff
@@ -91,6 +94,57 @@ public AdminUserController(DatabaseContext database)
...
+            .ExecuteUpdateAsync(s =>
+                s.SetProperty(c => c.Deleted, true)
+                 .SetProperty(c => c.DeletedBy, user.Username)
+                 .SetProperty(c => c.DeletedType, "moderator"));
+        Logger.Success($"Deleted comments for {targetedUser.Username} (id:{targetedUser.UserId})", LogArea.Admin);
+
+        await this.database.SendNotification(targetedUser.UserId,
+            "Your comments have been deleted by a moderator.");
+
+        await this.database.SaveChangesAsync();
```
Slendy, line 121:
```suggestion
```
ExecuteDelete and ExecuteUpdate don’t use change tracking so there’s no need to call SaveChanges

Slendy (commented):
I forgot in the previous review that batch updates and deletes don’t require you to call SaveChanges after

Slendy (approved):
lgtm

PR #1029, opened 2024-06-16
Base64 encode auth tokens

### LINKED ISSUE #1023: Email tokens not url encoded
**Describe the bug**
When emails are sent that involve tokens they should ideally be url encoded as that's what a url expects.

**To Reproduce**
1. Request any email that uses a token (password reset, or email verification)
2. look at the email and note the token on the end of the url, it's not url encoded

**Expected behavior**
Links should be either url encoded or only contain plain text

**Environment Details**
 - an email client, it's fairly universal.

Zaprit:
Base64 encode auth tokens to make them more URL compatible, i.e when they are used in emails.

Fixes #1023

### DIFF
--- a/ProjectLighthouse/Helpers/CryptoHelper.cs
+++ b/ProjectLighthouse/Helpers/CryptoHelper.cs
@@ -19,7 +19,14 @@ public static class CryptoHelper
     public static string GenerateAuthToken()
     {
         byte[] bytes = (byte[])GenerateRandomBytes(256);
-        
+
+        return BCryptHash(Sha256Hash(bytes));
+    }
+
+    public static string GenerateUrlToken()
+    {
+        byte[] bytes = (byte[])GenerateRandomBytes(256);
+
         return Convert.ToBase64String(Encoding.UTF8.GetBytes(BCryptHash(Sha256Hash(bytes))));
     }
 
--- a/ProjectLighthouse/Helpers/CryptoHelper.cs
+++ b/ProjectLighthouse/Helpers/CryptoHelper.cs
@@ -19,8 +19,8 @@ public static class CryptoHelper
     public static string GenerateAuthToken()
     {
         byte[] bytes = (byte[])GenerateRandomBytes(256);
-
-        return BCryptHash(Sha256Hash(bytes));
+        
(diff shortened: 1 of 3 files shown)


ProjectLighthouse/Helpers/CryptoHelper.cs
```diff
@@ -19,8 +19,8 @@ public static class CryptoHelper
     public static string GenerateAuthToken()
     {
         byte[] bytes = (byte[])GenerateRandomBytes(256);
-
-        return BCryptHash(Sha256Hash(bytes));
+
```
Slendy, line 22:
```suggestion

```
image
me when i see extra whitespace in a pr

Slendy (changes requested):

ProjectLighthouse/Helpers/CryptoHelper.cs
Zaprit: extra whitespace, bruh I meant to put that there :|

Zaprit: fine, but I'm not happy about it

ProjectLighthouse/Helpers/CryptoHelper.cs
```diff
@@ -19,8 +19,7 @@ public static class CryptoHelper
     public static string GenerateAuthToken()
     {
         byte[] bytes = (byte[])GenerateRandomBytes(256);
-
-        return BCryptHash(Sha256Hash(bytes));
+        return Convert.ToBase64String(Encoding.UTF8.GetBytes(BCryptHash(Sha256Hash(bytes))));
```
sudokoko, line 22: Will this interfere with the game at all, or is this logic separate from game authentication tokens?

HenryAsbridge: It will likely cause re-auths for game players, I'll do a quick check to make sure that it won't affect anything persistent

Zaprit: I've just made it so that only tokens in emails are base64 encoded

sudokoko (approved):

Slendy (approved):

PR #1035, opened 2024-06-19
Remove license check

### LINKED ISSUE #1032: "Potential License Violation"
I did not modify anything in the code...
Yet i get
```
Potential License Violation

This instance is a public-facing instance that has been modified without the changes published. You may be in violation of the [GNU Affero General Public License v3.0](https://github.com/LBPUnion/project-lighthouse/blob/main/LICENSE).

If you believe this is an error, please create an issue with the output of git status ran from the root of the server source code in the description on our [issue tracker](https://github.com/LBPUnion/project-lighthouse/issues).

If not, please publish the source code somewhere accessible to your users.
```

Here is the git status
```
On branch main
Your branch is up to date with 'origin/main'.

Untracked files:
  (use "git add <file>..." to include in what will be committed)
        data/

nothing added to commit but untracked files present (use "git add" to track)

```
Considering that if i would ever like to edit the website, where should i publish the changes?

jackcaver:
Closes #1032.

---

It's quite annoying and doesn't really do anything, it can also trigger on people who didn't modify the source code.
It is also very easy to disable it by making a commit in the local repo or just by changing `IsDirty` to always return `true`

### DIFF
--- a/ProjectLighthouse.Servers.Website/Pages/Debug/VersionInfoPage.cshtml
+++ b/ProjectLighthouse.Servers.Website/Pages/Debug/VersionInfoPage.cshtml
@@ -14,5 +14,4 @@
 <p><b>Branch</b>: @VersionHelper.Branch</p>
 <p><b>Build</b>: @VersionHelper.Build</p>
 <p><b>CommitHash</b>: @VersionHelper.CommitHash</p>
-<p><b>IsDirty</b>: @VersionHelper.IsDirty</p>
 <p><b>RepositoryUrl</b>: @VersionHelper.RepositoryUrl</p>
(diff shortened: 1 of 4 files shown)


sudokoko (changes requested):
https://github.com/LBPUnion/ProjectLighthouse/blob/be11e138f032ec306f9c732b4a3a3cee47740edc/ProjectLighthouse/Helpers/VersionHelper.cs#L22

This auto-property is unused and can be removed

jackcaver:
Technically it is still used by "version info" page, but sure, i can remove it, since that debug page only displays the value of this property

sudokoko:
Yes, that's fine. Just remove the part on the debug page too.

jackcaver:
ok, removed it from those 2 places, is there anything else i need to change?

sudokoko (approved):

Slendy (approved):

PR #1055, opened 2024-08-27
filter locked levels from lucky dip

Metraberryy:
filters locked levels out of lucky dip
iirc this was the functionality on official servers, and makes sense anyway

### DIFF
--- a/ProjectLighthouse.Servers.GameServer/Extensions/ControllerExtensions.cs
+++ b/ProjectLighthouse.Servers.GameServer/Extensions/ControllerExtensions.cs
@@ -175,6 +175,9 @@ public static class ControllerExtensions
 
         if (!queryBuilder.GetFilters(typeof(CrossControlFilter)).Any())
             queryBuilder.AddFilter(new ExcludeCrossControlFilter());
+        
+        if (controller.Request.Path.Value.EndsWith("lbp2luckydip"))
+            queryBuilder.AddFilter(new LockedSlotFilter());
 
         queryBuilder.AddFilter(new SubLevelFilter(token.UserId));
         queryBuilder.AddFilter(new HiddenSlotFilter());
--- a/ProjectLighthouse.Servers.GameServer/Extensions/ControllerExtensions.cs
+++ b/ProjectLighthouse.Servers.GameServer/Extensions/ControllerExtensions.cs
@@ -176,7 +176,7 @@ public static class ControllerExtensions
         if (!queryBuilder.GetFilters(typeof(CrossControlFilter)).Any())
             queryBuilder.AddFilter(new ExcludeCrossControlFilter());
         
-        if (controller.Request.Path.Value.EndsWith("lbp2luckydip"))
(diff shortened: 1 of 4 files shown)


ProjectLighthouse.Servers.GameServer/Extensions/ControllerExtensions.cs
```diff
@@ -175,6 +175,9 @@ void ParseLbp3Query(string key, Action allMust, Action noneCan, Action dontCare)
 
         if (!queryBuilder.GetFilters(typeof(CrossControlFilter)).Any())
             queryBuilder.AddFilter(new ExcludeCrossControlFilter());
+        
+        if (controller.Request.Path.Value.EndsWith("lbp2luckydip"))
```
sudokoko, line 179: Dereference of a possibly null reference

sudokoko (commented):
Warnings caught by the CI

sudokoko (approved):
This looks fine to me

ProjectLighthouse.Servers.GameServer/Extensions/ControllerExtensions.cs
```diff
@@ -175,6 +175,9 @@ void ParseLbp3Query(string key, Action allMust, Action noneCan, Action dontCare)
 
         if (!queryBuilder.GetFilters(typeof(CrossControlFilter)).Any())
             queryBuilder.AddFilter(new ExcludeCrossControlFilter());
+        
+        if (controller.Request.Path.Value != null && controller.Request.Path.Value.EndsWith("lbp2luckydip"))
+            queryBuilder.AddFilter(new LockedSlotFilter());
```
Slendy, line 180:
```suggestion
```
I think since this method is mostly for filtering based on the query params of the request, this would be better suited in the endpoint controllers for lucky dip. Specifically, in [LuckyDipSlots()](https://github.com/LBPUnion/ProjectLighthouse/blob/9ac74312ce44f2bc47eb447d81c305fc83d17e94/ProjectLighthouse.Servers.GameServer/Controllers/Slots/SlotsController.cs#L265) for LBP2 and in [LuckyDipCategory.cs](https://github.com/LBPUnion/ProjectLighthouse/blob/9ac74312ce44f2bc47eb447d81c305fc83d17e94/ProjectLighthouse.Servers.GameServer/Types/Categories/LuckyDipCategory.cs#L20) for LBP3. You should be able to do `queryBuilder.AddFilter(new LockedSlotFilter())` for both of them.

Slendy (changes requested):
I think that the code for excluding locked levels from lucky dip should be in the lucky dip specific functions rather than the general request param parsing function. Other than that, this looks good.

ProjectLighthouse.Servers.GameServer/Extensions/ControllerExtensions.cs
Metraberryy: for sure, this is the less jank implementation i was looking for ! for some reason it didnt occur to me that i could access the querybuilder outside of filterfromrequest

Slendy (approved):

PR #1060, opened 2024-09-04
Implement slot settings toggles for lock, shareable, sublevel, and LBP1-only. 

FeTetra:
This PR implements some checkbox UI in the slot settings for marking different slot flags accessible in-game. Locking, making shareable, making a sub level, and making a level LBP1-only, will be accessible on the website via the slot settings.


FeTetra:
Whether or not I made the checkboxes submit from the form correctly is entirely up to fate...

FeTetra:
I don't believe there's a case where a form submission could provide a null boolean from a checkbox, so I removed the tests for null booleans after I forgot when making the booleans not nullable.

Slendy:
Can you fix your commit history for this branch so it doesn’t have commits from your previous PR? In the future, you should probably make separate branches on your fork for each new pull request you plan to open.

FeTetra:
> Can you fix your commit history for this branch so it doesn’t have commits from your previous PR? In the future, you should probably make separate branches on your fork for each new pull request you plan to open.

Oops, that's my bad

PR #1082, opened 2024-11-14
Fix missing filtering, filter inconsistencies, and filter logging

FeTetra:
This PR deals with places where filtering should happen but does not. I have also moved logging into the filter helper itself to reduce duplicate code and remove inconsistencies where some filtered items are logged and some are not.

### DIFF
--- a/ProjectLighthouse.Servers.GameServer/Controllers/MessageController.cs
+++ b/ProjectLighthouse.Servers.GameServer/Controllers/MessageController.cs
@@ -145,7 +145,7 @@ along with this program.  If not, see <https://www.gnu.org/licenses/>.";
 
         if (ServerConfiguration.Instance.LogChatMessages) Logger.Info($"{username}: \"{message}\"", LogArea.Filter);
 
-        message = CensorHelper.FilterMessage(message,"in-game message", username);
+        message = CensorHelper.FilterMessage(message, "in-game message", username);
 
         return this.Ok(message);
     }
--- a/ProjectLighthouse.Servers.GameServer/Controllers/MessageController.cs
+++ b/ProjectLighthouse.Servers.GameServer/Controllers/MessageController.cs
@@ -146,7 +146,7 @@ along with this program.  If not, see <https://www.gnu.org/licenses/>.";
 
         if (ServerConfiguration.Instance.LogChatMessages) Logger.Info($"{username}: \"{message}\"", LogArea.Filter);
 
-        message = CensorHelper.FilterMessage(message, Location.ChatMessage, username);
+        message = CensorHelper.FilterMessage(message, FilterLocation.ChatMessage, username);
 
         return this.Ok(message);
     }
--- a/ProjectLighthouse.Servers.GameServer/Controllers/MessageController.cs
+++ b/ProjectLighthouse.Servers.GameServer/Controllers/MessageController.cs
@@ -143,15 +143,8 @@ along with this program.  If not, see <https://www.gnu.org/licenses/>.";
 
         string username = await this.database.UsernameFromGameToken(token);
 
-        string filteredText = CensorHelper.FilterMessage(message);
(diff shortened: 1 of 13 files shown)


ProjectLighthouse/Helpers/CensorHelper.cs
```diff
@@ -17,7 +19,7 @@ public static class CensorHelper
         "UwU", "OwO", "uwu", "owo", "o3o", ">.>", "*pounces on you*", "*boops*", "*baps*", ":P", "x3", "O_O", "xD", ":3", ";3", "^w^",
     };
 
-    public static string FilterMessage(string message)
+    public static string FilterMessage(string message, string filterLocation = null, string username = null)
```
sudokoko, line 22:
```suggestion
    public static string FilterMessage(string message, FilterLocation filterLocation = null, string username = null)
```
An enum would probably be better for the filter location, you'd have to create this yourself (don't commit this directly)

ProjectLighthouse.Servers.Website/Pages/Debug/FilterTestPage.cshtml.cs
```diff
@@ -19,7 +19,7 @@ public FilterTestPage(DatabaseContext database) : base(database)
     public IActionResult OnGet(string? text = null)
     {
         #if DEBUG
-        if (text != null) this.FilteredText = CensorHelper.FilterMessage(text);
+        if (text != null) this.FilteredText = CensorHelper.FilterMessage(text, "test filter", null);
```
sudokoko, line 22:
```suggestion
        if (text != null) this.FilteredText = CensorHelper.FilterMessage(text, "test filter");
```
The default value is already `null`

ProjectLighthouse.Servers.GameServer/Controllers/MessageController.cs
```diff
@@ -143,15 +143,10 @@ public async Task<IActionResult> Filter(IMailService mailService)
...
 
-        string filteredText = CensorHelper.FilterMessage(message);
-
         if (ServerConfiguration.Instance.LogChatMessages) Logger.Info($"{username}: \"{message}\"", LogArea.Filter);
 
-        if (ServerConfiguration.Instance.LogChatFiltering && filteredText != message)
-            Logger.Info(
-                $"Censored profane word(s) from in-game text sent by {username}: \"{message}\" => \"{filteredText}\"",
-                LogArea.Filter);
+        message = CensorHelper.FilterMessage(message,"in-game message", username);
```
sudokoko, line 148:
```suggestion
        message = CensorHelper.FilterMessage(message, "in-game message", username);
```

sudokoko (changes requested):

ProjectLighthouse.Servers.Website/Pages/Debug/FilterTestPage.cshtml.cs
FeTetra: Admittedly I considered changing this but I just lazily didn't because there are already like 3 warnings in the tests that haven't been fixed for ages :clueless:.

ProjectLighthouse/Helpers/CensorHelper.cs
FeTetra: I agree, I felt like using a string was kinda janky. I'll fix this tomorrow.

sudokoko (approved):

ProjectLighthouse/Types/Filter/FilterLocation.cs
```diff
@@ -0,0 +1,13 @@
+namespace LBPUnion.ProjectLighthouse.Types.Filter;
+
+public enum Location
```
Slendy, line 3:
```suggestion
public enum FilterLocation
```
I think this should have a more specific name otherwise it could be confused with the Location used for users and levels.

Slendy (changes requested):

Slendy (approved):

PR #1089, opened 2025-04-16
Handle long entitlements section in NPTicket 2.1 tickets

Zaprit:
This PR correctly reads the section header of the "entitlements" section within the NPTicket so it can skip the right number of bytes, as in rare cases there is extra data in this section.


daggintosh (approved):

m88youngling:
zaprit you need to bludgeon qodana again

PR #1091, opened 2025-05-26
Implement refresh autodiscover

Zaprit:
This PR adds the refresh autodiscover endpoint to the website server, so that refresher can be used with minimal friction

### DIFF
--- a/ProjectLighthouse.Servers.Website/Controllers/AutoDiscoverController.cs
+++ b/ProjectLighthouse.Servers.Website/Controllers/AutoDiscoverController.cs
@@ -1,0 +1,26 @@
+using LBPUnion.ProjectLighthouse.Configuration;
+using LBPUnion.ProjectLighthouse.Servers.Website.Types;
+using Microsoft.AspNetCore.Mvc;
+
+namespace LBPUnion.ProjectLighthouse.Servers.Website.Controllers;
+
+[ApiController]
+public class AutoDiscoverController: ControllerBase
+{
+    [ResponseCache(Duration = 86400)]
+    [HttpGet("/autodiscover")]
+    [Produces("application/json")]
+    public IActionResult AutoDiscover()
+    {
+        AutoDiscoverResponse resp = new()
+        {
+            Version = 3,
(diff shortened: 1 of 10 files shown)


ProjectLighthouse.Servers.Website/Controllers/AutoDiscoverController.cs
```diff
@@ -0,0 +1,26 @@
...
+using LBPUnion.ProjectLighthouse.Servers.Website.Types;
+using Microsoft.AspNetCore.Mvc;
+
+namespace LBPUnion.ProjectLighthouse.Servers.Website.Controllers;
+
+[ApiController]
+public class AutoDiscoverController: ControllerBase
+{
+    [ResponseCache(Duration = 86400)]
+    [HttpGet("/autodiscover")]
```
Slendy, line 11:
```suggestion
    [HttpGet("autodiscover")]
```
Doesn't affect functionality but standard is to leave off the leading slash

ProjectLighthouse.Servers.Website/Controllers/AutoDiscoverController.cs
```diff
@@ -0,0 +1,26 @@
+using LBPUnion.ProjectLighthouse.Configuration;
+using LBPUnion.ProjectLighthouse.Servers.Website.Types;
+using Microsoft.AspNetCore.Mvc;
+
+namespace LBPUnion.ProjectLighthouse.Servers.Website.Controllers;
+
+[ApiController]
+public class AutoDiscoverController: ControllerBase
+{
+    [ResponseCache(Duration = 86400)]
```
Slendy, line 10: Did you test if this actually does anything? I think in ASP.NET you have to explicitly enable response caching at startup.

Slendy (approved):

FeTetra (approved):
Nothing stands out to me.

ProjectLighthouse.Servers.Website/Controllers/AutoDiscoverController.cs
Zaprit: image
yep, it works fine

PR #1094, opened 2025-07-20
Add command to remove user's 2fa configuration.

Zaprit:
This PR adds a command to remove a users 2fa configuration, so that the database doesn't need to be manually edited.


kylometra (approved):
my only complaint is that the collection expression for aliases is inconsistent with the rest of the commands
i do like it but i think if it's gonna be changed it could be done in another pr; i wouldn't be too bothered if you keep it and change the rest in another pr though. lgtm either way

PR #1100, opened 2025-10-28
don't separate in-game search into keywords

kylometra:
it was basically impossible to search for anything with common words
hopefully this will be more usable


sudokoko (approved):
lgtm

Slendy (approved):

PR #1107, opened 2026-05-07
Remove Patchwork regex end

### LINKED ISSUE #1095: LBP Vita does not work because of User-Agent bug
**Describe the bug**
There is no offically supported way of patching LBP Vita as of right now, but for the future (and with LittleBigPatcher)

When you replace the User-Agent `MM CHTTPClient LBPVita %s` with `PatchworkLBPV 1.0`, and then try to connect, even though the `%s` is removed, the User-Agent will be `PatchworkLBPV 1.0 libhttp/3.74 (PS Vita)`, and the [regex](https://github.com/LBPUnion/ProjectLighthouse/blob/aeba706391746f1516b19d1ba0122b12664fe7e9/ProjectLighthouse.Servers.GameServer/Helpers/PatchworkHelper.cs#L11) only allows a full match of `PatchworkLBPV 1.0`

This is an unintended way of blocking LBP Vita connections, perhaps the code can be fixed (here is a [correct implementation](https://github.com/LittleBigRefresh/Refresh/blob/45976ba98101b8c9eeb3e99d58e51e577bc0477a/Refresh.Common/Extensions/RequestContextExtensions.cs#L43-L86))

And then LBP Vita can be explicitly blocked

**Game(s) affected**
LBP Vita

FeTetra:
It turns out this is actually a good idea and wont break anything.

Fixes #1095

### DIFF
--- a/ProjectLighthouse.Servers.GameServer/Helpers/PatchworkHelper.cs
+++ b/ProjectLighthouse.Servers.GameServer/Helpers/PatchworkHelper.cs
@@ -8,7 +8,7 @@ public static partial class PatchworkHelper
     private static readonly int requiredMajor = ServerConfiguration.Instance.Authentication.PatchworkMajorVersionMinimum;
     private static readonly int requiredMinor = ServerConfiguration.Instance.Authentication.PatchworkMinorVersionMinimum;
 
-    [GeneratedRegex(@"^PatchworkLBP[123V] (\d{1,5})\.(\d{1,5})$")]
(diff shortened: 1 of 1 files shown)


jackcaver:
please merge this

Zaprit (approved):

sudokoko (approved):
lgtm

jackcaver:
so... is it getting merged?

Zaprit:
I forgor 💀

PR #1109, opened 2026-08-12
Added 3 New LBP3 Categories

W0lf4llo:
I added support for LBP3's recently played category.

Essentially, LBP2 tracks which levels you've played client side, and just reports it back whenever you click on "Recently Played". LBP3 on the original servers would handle all of this server sided. This essentially works by recording the recently played community levels via the games "gameState" request, then it stores the recently played slot ID's and the timestamps per each user. I set a limit so that it keeps a maximum of 20 recently played levels, when you replay a level that was already on the list, it changes the timestamp to have it re-appear at the top of the list. The game also only applies to users who are playing on LBP3. If you play LBP1 or LBP2, it will not change this list whatsoever. 

image


W0lf4llo:
I wrote a fairly simple recommendation system for the server which is VERY similar to how Sony used to handle it. I left my comments in there so you can easily read what everything does and why its there. If anything else is needed to change in this PR please let me know! I just figured I'd stack it on top of this pr as i built this on top of my already existing work.

Toastbrot236:
Is there a reason you're using `/gameState` instead of `/play` to find out what levels a user has played?

W0lf4llo:
Also, I forgot to thank you on here for the recommendation, as using /play was a far better way of tracking when a user actually enters a community level. I appreciate it!

W0lf4llo:
Also, this is my last commit to the PR unless an issue has been found.. which I'm sure there will 100% be issues, just feel free to let me know, and ask any questions if needed!

ProjectLighthouse.Servers.GameServer/Controllers/Matching/MatchController.cs
```diff
@@ -14,6 +14,7 @@
 using Microsoft.AspNetCore.Authorization;
 using Microsoft.AspNetCore.Mvc;
 using Microsoft.EntityFrameworkCore;
+using LBPUnion.ProjectLighthouse.Types.Users;
```
FeTetra:
```suggestion
```

This can be removed.

ProjectLighthouse.Servers.GameServer/Controllers/Matching/MatchController.cs
```diff
@@ -32,7 +33,10 @@ public MatchController(DatabaseContext database)
 
     [HttpPost("gameState")]
     [Produces("text/plain")]
-    public IActionResult GameState() => this.Ok("VALID");
+    public IActionResult GameState()
+    {
+        return this.Ok("VALID");
+    }
```
FeTetra:
```suggestion
    public IActionResult GameState() => this.Ok("VALID");
```
Why was this reformatted?

ProjectLighthouse.Servers.GameServer/Types/Categories/NewestLevelsCategory.cs
```diff
@@ -10,13 +10,17 @@ namespace LBPUnion.ProjectLighthouse.Servers.GameServer.Types.Categories;
...
-    public override string Tag => "newest";
+  public override string Name { get; set; } = "Newest Levels";
+  public override string Description { get; set; } = "The most recently published content";
+  public override string IconHash { get; set; } = "g820623";
+  public override string Endpoint { get; set; } = "newest";
+  public override string Tag => "newest";
+  public override string[] Sorts { get; } =
+  {
+      "date"
+    };
```
FeTetra:
```suggestion
    public override string Name { get; set; } = "Newest Levels";
    public override string Description { get; set; } = "The most recently published content";
    public override string IconHash { get; set; } = "g820623";
    public override string Endpoint { get; set; } = "newest";
    public override string Tag => "newest";
    public override string[] Sorts { get; } =
    {
        "date"
    };
```

ProjectLighthouse.Servers.GameServer/Types/Categories/NewestLevelsCategory.cs
```diff
@@ -10,13 +10,17 @@ namespace LBPUnion.ProjectLighthouse.Servers.GameServer.Types.Categories;
...
 
-    public override IQueryable<SlotEntity> GetItems(DatabaseContext database, GameTokenEntity token, SlotQueryBuilder queryBuilder) =>
-        database.Slots.Where(queryBuilder.Build())
-            .ApplyOrdering(new SlotSortBuilder<SlotEntity>().AddSort(new FirstUploadedSort()));
+  public override IQueryable<SlotEntity> GetItems(DatabaseContext database, GameTokenEntity token, SlotQueryBuilder queryBuilder) =>
+      database.Slots.Where(queryBuilder.Build())
+          .ApplyOrdering(new SlotSortBuilder<SlotEntity>().AddSort(new FirstUploadedSort()));
```
FeTetra:
```suggestion
    public override IQueryable<SlotEntity> GetItems(DatabaseContext database, GameTokenEntity token, SlotQueryBuilder queryBuilder) =>
        database.Slots.Where(queryBuilder.Build())
            .ApplyOrdering(new SlotSortBuilder<SlotEntity>().AddSort(new FirstUploadedSort()));
```

ProjectLighthouse.Servers.GameServer/Types/Categories/RecommendedCategory.cs
FeTetra, line 1: This file has tons of formatting issues, there are private fields at the top which should all be configuration options, you should consider making a new config file to store search and category query options. There are also a lot of db operations which could potentially become very expensive, especially if the number of things such as neighbors can potentially approach the set max neighbor value of 250. A lot of the variable naming choices are confusing, such as the repeated use of `h` in lambda expressions which can seemingly refer to a hearted user or a hearted slot of which can either be hearted by you, a user you hearted, or (in the case of only slots) a distinct user who hearted the same slot as the user you hearted. In every other case with categories, they directly return an `IQueryable` for the database, in this case, you do all the db work here, and then convert the list of results into an `IQueryable`. This sidesteps the purpose of returning an `IQueryable` to begin with, as it is intended to only provide the database query for code within the list controller to actually perform the returned query alongside other things like pagination. The code will always attempt to grab a maximum of 1000 slot candidates for the resulting list, making the later query pagination in the code totally pointless.

FeTetra (changes requested):
There is a lot more to review, but I'm going to leave this for now.

ProjectLighthouse/Database/DatabaseContext.Slots.cs
```diff
@@ -87,4 +90,59 @@ public async Task UnqueueLevel(int userId, SlotEntity queuedSlot)
         await this.SaveChangesAsync();
     }
 
-}
\ No newline at end of file
+    public async Task RecordRecentlyPlayedLevel(int userId, int slotId, bool saveChanges = true)
+    {
+        RecentlyPlayedEntity? recentlyPlayed = await this.RecentlyPlayed.FirstOrDefaultAsync(r => r.UserId == userId);
+
+        long now = DateTimeOffset.UtcNow.ToUnixTimeMilliseconds();
```
Slendy, line 97:
```suggestion
        long now = TimeHelper.TimestampMillis;
```

ProjectLighthouse/Database/DatabaseContext.Slots.cs
```diff
@@ -87,4 +90,59 @@ public async Task UnqueueLevel(int userId, SlotEntity queuedSlot)
         await this.SaveChangesAsync();
     }
 
-}
\ No newline at end of file
+    public async Task RecordRecentlyPlayedLevel(int userId, int slotId, bool saveChanges = true)
```
Slendy: This `saveChanges` parameter is only ever passed as false, I don't think it's necessary.

ProjectLighthouse/Migrations/20260812004441_AddRecentlyPlayed.Designer.cs
```diff
@@ -0,0 +1,1602 @@
+﻿// <auto-generated />
```
Slendy, line 1: This entire file doesn't need to be included and can be deleted

ProjectLighthouse/Types/Levels/CategoryDefaults.cs
```diff
@@ -0,0 +1,32 @@
+#nullable enable
```
Slendy: This should be in the serialization namespace/folder

ProjectLighthouse/Types/Levels/CategoryDefaults.cs
```diff
@@ -0,0 +1,32 @@
...
+    public bool ShouldSerializeGameFilter() => !string.IsNullOrWhiteSpace(this.GameFilter);
+
+    public bool ShouldSerializeDateFilterType() => !string.IsNullOrWhiteSpace(this.DateFilterType);
+
+    public bool ShouldSerializeIncludePlayed() => this.IncludePlayed.HasValue;
+
+    public bool ShouldSerializeTeamPicked() => !string.IsNullOrWhiteSpace(this.TeamPicked);
+
+    public bool ShouldSerializeBlacklisted() => !string.IsNullOrWhiteSpace(this.Blacklisted);
+}
```
Slendy, line 32: Instead of a bunch of ShouldSerialize you should use 
```c#
[DefaultValue("")]
```
on any nullable fields that you don't want to show up if they are empty. If the field is an object then you can use `null` instead of the empty double quotes

example:
```c#
    [DefaultValue("")]
    [XmlElement("gameFilter")]
    public string? GameFilter { get; set; }
```

There's a couple places that this should be fixed but I'm not going to comment on all of them for brevity.

```suggestion
```

ProjectLighthouse/Types/Serialization/GameCategory.cs
```diff
@@ -9,46 +11,66 @@ public class GameCategory : ILbpSerializable
...
-    public string[] Types { get; set; }
+    public string[] Types { get; set; } = [];
 
-    [XmlElement("tag")]
-    public string Tag { get; set; }
+    //This will likely be used in the future if Companion Capers ever get added in LBP3
+    [XmlElement("param")]
+    public string? Param { get; set; }
+
+    public bool ShouldSerializeParam() => !string.IsNullOrWhiteSpace(this.Param);
```
Slendy:
```suggestion
```

ProjectLighthouse/Types/Serialization/GameCategory.cs
```diff
@@ -9,46 +11,66 @@ public class GameCategory : ILbpSerializable
...
+
+    [XmlElement("defaultFilters")]
+    public CategoryDefaults? DefaultFilters { get; set; }
+
+    public bool ShouldSerializeDefaultFilters() => this.DefaultFilters is not null;
```
Slendy:
```suggestion
```

ProjectLighthouse.Servers.GameServer/Controllers/Matching/EnterLevelController.cs
```diff
@@ -89,6 +89,11 @@ public async Task<IActionResult> PlayLevel(string slotType, int slotId)
                 return this.BadRequest();
         }
 
+        if (token.GameVersion == GameVersion.LittleBigPlanet3 && slotType == "user")
```
Slendy, line 92: The slotType check isn't necessary because story levels are excluded via the check on line 36
```suggestion
        if (token.GameVersion == GameVersion.LittleBigPlanet3)
```

ProjectLighthouse.Servers.GameServer/Types/Categories/RecommendedCategory.cs
```diff
@@ -0,0 +1,192 @@
...
+    {
+        public SlotEntity Slot { get; set; } = null!;
+        public int SearchScore { get; set; }
+        public int PrevSearchScore { get; set; }
+        public int Hearts { get; set; }
+        public int Likes { get; set; }
+    }
+
+    public IQueryable<ScoredSlot> GetScoredItems(DatabaseContext database, GameTokenEntity token, SlotQueryBuilder queryBuilder)
+    {
```
Slendy, line 34: I think this should be refactored into 2 separate functions for actually calculating the user's search scores, and then fetching the levels based on the scores should take place here.

ProjectLighthouse.Servers.GameServer/Types/Categories/RecommendedCategory.cs
```diff
@@ -0,0 +1,192 @@
...
+
+    public static ILbpSerializable CreateSerializableSlot(ScoredSlot recommendation, GameTokenEntity token)
+    {
+        SlotBase serialized = SlotBase.CreateFromEntity(recommendation.Slot, token);
+
+        if (serialized is GameUserSlot userSlot)
+        {
+            userSlot.SearchScore = recommendation.SearchScore;
+            userSlot.PrevSearchScore = recommendation.PrevSearchScore;
+        }
```
Slendy, line 173: Are these fields actually parsed by the game or is it just something that the official servers did?

ProjectLighthouse/Database/DatabaseContext.Slots.cs
Slendy, line 93: This `saveChanges` parameter is only ever passed as false, I don't think it's necessary.

ProjectLighthouse/Configuration/CategoryConfiguration.cs
```diff
@@ -0,0 +1,56 @@
...
+    // THIS VALUE MUST BE INCREMENTED FOR EVERY CONFIG CHANGE!
+    //
+    // This is so Lighthouse can properly identify outdated configurations and update them with newer settings accordingly.
+    // If you are modifying anything here, this value MUST be incremented.
+    // Thanks for listening~
+    public override int ConfigVersion { get; set; } = 1;
+    public override string ConfigName { get; set; } = "CategoryConfig.yml";
+    public override bool NeedsConfiguration { get; set; } = false;
+
+    public List<string> OrderOfCategory { get; set; } = new()
```
Slendy, line 19: `OrderOfCategory` feels like a somewhat misleading name since removing a category from here prevents it from showing up in-game. `Categories` might fit better but I'm open to suggestions.

ProjectLighthouse/Configuration/CategoryConfiguration.cs
```diff
@@ -0,0 +1,56 @@
...
+        "most_played",
+        "my_playlists",
+        "favourite_creators",
+        "queue",
+        "hearted_levels",
+        "highest_rated",
+        "lucky_dip",
+    };
+
+    public List<string> DisabledCategories { get; set; } = new()
```
Slendy, line 36: I don't know if having a separate field for disabled categories is necessary since you can disable categories by removing them form the category list above. In practice you would just have to comment out a line in the config

ProjectLighthouse/Database/DatabaseContext.Slots.cs
```diff
@@ -87,4 +90,59 @@ public async Task UnqueueLevel(int userId, SlotEntity queuedSlot)
...
+
+            if (existingIndex < recentlyPlayed.LastPlayedAt.Count)
+                recentlyPlayed.LastPlayedAt.RemoveAt(existingIndex);
+        }
+
+        //The most recently played level is at the top
+        recentlyPlayed.SlotIds.Insert(0, slotId);
+        recentlyPlayed.LastPlayedAt.Insert(0, now);
+
+        //Max of 30 levels
```
Slendy, line 135: This should use the MaxLevels config value

ProjectLighthouse/Database/DatabaseContext.Slots.cs
```diff
@@ -87,4 +90,59 @@ public async Task UnqueueLevel(int userId, SlotEntity queuedSlot)
...
+        if (recentlyPlayed == null)
+        {
+            this.RecentlyPlayed.Add(new RecentlyPlayedEntity
+            {
+                UserId = userId,
+                SlotIds = new List<int> { slotId },
+                LastPlayedAt = new List<long> { now },
+            });
+
+            if (saveChanges)await this.SaveChangesAsync();
```
Slendy, line 109:
```suggestion
            if (saveChanges) await this.SaveChangesAsync();
```

ProjectLighthouse/Database/DatabaseContext.Slots.cs
```diff
@@ -87,4 +90,59 @@ public async Task UnqueueLevel(int userId, SlotEntity queuedSlot)
...
+        long now = DateTimeOffset.UtcNow.ToUnixTimeMilliseconds();
+
+        //Level at the top of the recently played category for the user
+        if (recentlyPlayed == null)
+        {
+            this.RecentlyPlayed.Add(new RecentlyPlayedEntity
+            {
+                UserId = userId,
+                SlotIds = new List<int> { slotId },
+                LastPlayedAt = new List<long> { now },
```
Slendy, line 106:
```suggestion
                SlotIds = [slotId,],
                LastPlayedAt = [now,],
```

ProjectLighthouse/Database/DatabaseContext.Slots.cs
```diff
@@ -87,4 +90,59 @@ public async Task UnqueueLevel(int userId, SlotEntity queuedSlot)
...
 
-}
\ No newline at end of file
+    public async Task RecordRecentlyPlayedLevel(int userId, int slotId, bool saveChanges = true)
+    {
+        RecentlyPlayedEntity? recentlyPlayed = await this.RecentlyPlayed.FirstOrDefaultAsync(r => r.UserId == userId);
+
+        long now = DateTimeOffset.UtcNow.ToUnixTimeMilliseconds();
+
+        //Level at the top of the recently played category for the user
```
Slendy, line 99: This comment doesn't really make sense, is it supposed to go somewhere else?

ProjectLighthouse.Servers.GameServer/Types/Categories/HighestRatedCategory.cs
```diff
@@ -4,23 +4,33 @@
...
     public override IQueryable<SlotEntity> GetItems(DatabaseContext database, GameTokenEntity token, SlotQueryBuilder queryBuilder) =>
         database.Slots.Select(s => new SlotMetadata
-            {
-                Slot = s,
-                ThumbsUp = database.RatedLevels.Count(r => r.SlotId == s.SlotId && r.Rating == 1),
-            })
+        {
+            Slot = s,
+            ThumbsUp = database.RatedLevels.Count(r => r.SlotId == s.SlotId && r.Rating == 1),
+        })
```
Slendy, line 33:
```suggestion
            {
                Slot = s,
                ThumbsUp = database.RatedLevels.Count(r => r.SlotId == s.SlotId && r.Rating == 1),
            })
```
Fix indentation here

ProjectLighthouse.Servers.GameServer/Types/Categories/BusiestCategory.cs
```diff
@@ -0,0 +1,53 @@
...
+{
+    public override string Name { get; set; } = "Busiest";
+    public override string Description { get; set; } = "Levels being played right now!";
+    public override string IconHash { get; set; } = "g820602";
+    public override string Endpoint { get; set; } = "busiest";
+    public override string Tag => "busiest";
+    public override string[] Sorts { get; } =
+    {
+      "relevance",
+    };
```
Slendy, line 21:
```suggestion
    public override string[] Sorts { get; } = ["relevance",];
```
This change should be implemented in all Category classes where possible

ProjectLighthouse.Servers.GameServer/Types/Categories/MostHeartedCategory.cs
```diff
@@ -16,13 +17,24 @@ public class MostHeartedCategory : SlotCategory
...
     public override IQueryable<SlotEntity> GetItems(DatabaseContext database, GameTokenEntity token, SlotQueryBuilder queryBuilder) =>
         database.Slots.Select(s => new SlotMetadata
-            {
-                Slot = s,
-                Hearts = database.HeartedLevels.Count(r => r.SlotId == s.SlotId),
-            })
+        {
+            Slot = s,
+            Hearts = database.HeartedLevels.Count(r => r.SlotId == s.SlotId),
+        })
```
Slendy, line 37:
```suggestion
            {
                Slot = s,
                Hearts = database.HeartedLevels.Count(r => r.SlotId == s.SlotId),
            })
```
Fix indentation

ProjectLighthouse.Servers.GameServer/Types/Categories/RecentlyPlayedCategory.cs
```diff
@@ -0,0 +1,76 @@
...
+    {
+        RecentlyPlayedEntity? recentlyPlayed = database.RecentlyPlayed
+            .AsNoTracking()
+            .FirstOrDefault(r => r.UserId == token.UserId);
+
+        if (recentlyPlayed == null || recentlyPlayed.SlotIds.Count == 0)
+            return database.Slots.Where(_ => false);
+
+        List<int> slotIds = recentlyPlayed.SlotIds
+            .Take(30)
```
Slendy, line 42: This should also use the config value instead of hardcoding it

Slendy (changes requested):
Thanks for your contribution!

Here are a few more things that should be addressed that aren't covered in the requested changes:
* Comments should have a space between the two backslashes and the first character of the comment (`//comment goes here` should be `// comment goes here`)
* Indentation of code and comments in certain places is off

This could honestly be split into separate PRs since from what I can tell the recently played and recommended categories have nothing to do with each other, but it's fine given how dead development is now.

ProjectLighthouse.Servers.GameServer/Types/Categories/RecentlyPlayedCategory.cs
```diff
@@ -0,0 +1,76 @@
...
+      "relevance",
+    };
+
+    public override IQueryable<SlotEntity> GetItems(
+        DatabaseContext database,
+        GameTokenEntity token,
+        SlotQueryBuilder queryBuilder)
+    {
+        RecentlyPlayedEntity? recentlyPlayed = database.RecentlyPlayed
+            .AsNoTracking()
```
sudokoko, line 35: You're using `.AsNoTracking()` in some places and not others, and additionally - I don't believe we've really ever marked database calls as explicitly untracked.

```suggestion
```

ProjectLighthouse/Configuration/CategoryConfiguration.cs
```diff
@@ -0,0 +1,56 @@
...
+    public RecommendedCategoryConfig Recommended { get; set; } = new();
+    public RecentlyPlayedConfig RecentlyPlayed { get; set; } = new();
+    public override ConfigurationBase<CategoryConfiguration> Deserialize(IDeserializer deserializer, string text) => deserializer.Deserialize<CategoryConfiguration>(text);
+}
+
+public class RecommendedCategoryConfig
+{
+    public int MaxNeighbors { get; set; } = 250;
+    public int MinimumNeighborOverlap { get; set; } = 2;
+    public int MaxCandidatePool { get; set; } = 1000;
```
sudokoko, line 50:
```suggestion
```

From what I can tell, this isn't used anywhere. I don't know if you intended for it to be.

ProjectLighthouse.Servers.GameServer/Types/Categories/RecommendedCategory.cs
```diff
@@ -0,0 +1,192 @@
...
+
+            join score in scores
+                on slot.SlotId equals score.SlotId
+
+            let hearts = database.HeartedLevels.Count(heartedLevel =>
+                heartedLevel.SlotId == slot.SlotId)
+
+            let likes = database.RatedLevels.Count(rating =>
+                rating.SlotId == slot.SlotId &&
+                rating.Rating == 1)
```
sudokoko, line 140: This query seems to do a lot of unbounded work and performs multiple COUNT operations per slot. Could have performance issues at scale.

Perhaps this is where you intended to use `CategoryConfiguration.Instance.Recommended.MaxCandidatePool` to limit the candidate set before performing expensive aggregates?

Could also pre-aggregate the likes and hearts in separate grouped queries instead of calling COUNT inside of the projection.

`.Take(config.MaxCandidatePool)`

ProjectLighthouse/Types/Serialization/GameCategory.cs
```diff
@@ -9,46 +11,66 @@ public class GameCategory : ILbpSerializable
...
     [DefaultValue("")]
-    public string Description { get; set; }
+    public string Description { get; set; } = string.Empty;
 
     [XmlElement("url")]
-    public string Url { get; set; }
+    public string Url { get; set; } = string.Empty;
+
+    [XmlElement("tag")]
+    public string Tag { get; set; } = string.Empty;
```
sudokoko:
```suggestion
    [XmlElement("tag")]
    [DefaultValue("")]
    public string Tag { get; set; } = string.Empty;
```

You should assign DefaultValue attributes to the fields you added to remain consistent with the rest of the model.

ProjectLighthouse/Types/Serialization/GameUserSlot.cs
```diff
@@ -294,7 +303,11 @@ public async Task PrepareSerialization(DatabaseContext database)
         }
         #nullable disable
 
-        this.PlayerCount = RoomHelper.Rooms.Count(r => r.Slot.SlotType == SlotType.User && r.Slot.SlotId == this.SlotId);
+        Dictionary<int, int> playerCounts = RoomHelper.GetUserLevelPlayerCounts();
```
sudokoko, line 306: If PrepareSerialization is invoked for multiple slots in parallel, this could result in a performance issue with how this method is implemented. See comment in `GameUserSlot.cs`.

ProjectLighthouse/Helpers/RoomHelper.cs
```diff
@@ -156,6 +156,26 @@ public static Room CreateRoom(List<int> users, GameVersion roomVersion, Platform
         return Rooms.FirstOrDefault(room => room.PlayerIds.Any(p => p == userId));
     }
 
+    public static Dictionary<int, int> GetUserLevelPlayerCounts()
+    {
+        lock (RoomLock)
```
sudokoko, line 161: Continued from the comment on GameUserSlot.cs - method acquires a lock and builds a full dictionary each time. Consider caching the counts for the duration of the request or exposing an API where the caller computes the counts once and reuses them for multiple calls.

sudokoko (changes requested):
A couple things I noticed, on top of Slendy's review.

ProjectLighthouse.Servers.GameServer/Types/Categories/BusiestCategory.cs
```diff
@@ -0,0 +1,49 @@
...
+
+        ParameterExpression slotParameter = Expression.Parameter(typeof(SlotEntity), "slot");
+
+        MemberExpression slotIdProperty = Expression.Property(slotParameter, nameof(SlotEntity.SlotId));
+
+        Expression playerCountExpression = Expression.Constant(0);
+
+        foreach (KeyValuePair<int, int> playerCount in playerCounts)
+        {
+            playerCountExpression = Expression.Condition(Expression.Equal(slotIdProperty, Expression.Constant(playerCount.Key)), Expression.Constant(playerCount.Value), playerCountExpression);
```
FeTetra: Line is a bit long.
```suggestion
            playerCountExpression = Expression.Condition(
                Expression.Equal(slotIdProperty, Expression.Constant(playerCount.Key)),
                Expression.Constant(playerCount.Value), 
                playerCountExpression);
```

ProjectLighthouse.Servers.GameServer/Types/Categories/CategoryHelper.cs
FeTetra: Overall, I would argue a configuration to manually enable or disable categories by name is a bit unnecessary. But if others disagree with my sentiment, I still think it would be better to use the category's tag or something to put into the configuration rather than separate hardcoded names.

ProjectLighthouse/Configuration/CategoryConfiguration.cs
FeTetra: See my comment on `ProjectLighthouse.Servers.GameServer/Types/Categories/CategoryHelper.cs`

ProjectLighthouse.Servers.GameServer/Types/Categories/LuckyDipCategory.cs
```diff
@@ -15,12 +15,16 @@ public class LuckyDipCategory : SlotCategory
     public override string Description { get; set; } = "A random selection of content";
     public override string IconHash { get; set; } = "g820605";
     public override string Endpoint { get; set; } = "lucky_dip";
-    public override string Tag => "lucky_dip";
+    public override string Tag => "level_of_the_day";
```
FeTetra: Was the previous tag incorrect? Was nothing being broken before?

ProjectLighthouse.Servers.GameServer/Types/Categories/RecentlyPlayedCategory.cs
```diff
@@ -0,0 +1,32 @@
...
+    {
+        return (
+            from recentlyPlayed in database.RecentlyPlayed
+            join slot in database.Slots.Where(queryBuilder.Build())
+                on recentlyPlayed.SlotId equals slot.SlotId
+            where recentlyPlayed.UserId == token.UserId
+            orderby recentlyPlayed.LastPlayedAt descending
+            select slot
+        ).Take(CategoryConfiguration.Instance.RecentlyPlayed.MaxLevels);
+    }
```
FeTetra: This should be consistent with how it's done elsewhere in the code. This should work, but I'd recommend checking before merging the suggestion.
```suggestion
    public override IQueryable<SlotEntity> GetItems(DatabaseContext database, GameTokenEntity token, SlotQueryBuilder queryBuilder) => 
        database.RecentlyPlayed.Where(r => r.UserId == token.UserId)
            .OrderByDescending(r => r.LastPlayedAt)
            .Select(r => r.Slot)
            .Where(queryBuilder.Build())
            .Take(CategoryConfiguration.Instance.RecentlyPlayed.MaxLevels);
```

ProjectLighthouse/Migrations/DatabaseContextModelSnapshot.cs
FeTetra: Delete.

PR #1114, opened 2026-08-27
Implement Patchwork 2.0 user agent parsing

FeTetra:
This commit adds some code to `PatchworkHelper.cs` to check for an optional part of the user agent provided by a Patchwork client which informs the server that the client has a join key set. This allows us to optionally provide access to `/match` if a user does not have a join key set. There is a new option in `MatchmakingConfiguration.cs` called `PatchworkMatchmakingEnabled`, which if enabled, will override `MatchmakingEnabled` and allow matchmaking even if it is disabled on the server. Patchwork clients can still matchmake even if server matchmaking is completely enabled, but they will not be allowed to matchmake if a join key is set due to it breaking lobby connections.


ProjectLighthouse.Tests.GameApiTests/Unit/PatchworkUserAgentTests.cs
```diff
@@ -32,12 +32,12 @@ public void CanValidatePatchworkUserAgents()
         bool result;
         foreach (string userAgent in validUserAgents)
         {
-            result = PatchworkHelper.IsValidPatchworkUserAgent(userAgent);
+            result = PatchworkHelper.IsValidPatchworkUserAgent(userAgent, out _, out _, out _);
```
Slendy: Can you add some asserts for checking the out parameters of IsValidPatchworkUserAgent?

ProjectLighthouse/Migrations/20260827022735_AddPatchworkSessionDataToGameToken.Designer.cs
```diff
@@ -0,0 +1,1573 @@
+﻿// <auto-generated />
```
Slendy: This file doesn't need to be included and can be deleted

ProjectLighthouse/Types/Entities/Token/GameTokenEntity.cs
```diff
@@ -30,4 +30,9 @@ public class GameTokenEntity
     public string LocationHash { get; set; }
 
     public DateTime ExpiresAt { get; set; }
+
+    // Patchwork session data
+    public int? PatchworkMajor { get; set; } = 0;
+    public int? PatchworkMinor { get; set; } = 0;
+    public bool? PatchworkJoinKeyEnabled { get; set; } = false;
```
Slendy: What's the purpose of storing the major and minor? You never use them after setting them during login. 

Also, what's the advantage of storing this info in the token rather than parsing the user agent for each request?

ProjectLighthouse.Servers.GameServer/Controllers/Matching/MatchController.cs
```diff
@@ -46,7 +46,12 @@ public async Task<IActionResult> Match()
         await LastContactHelper.SetLastContact(this.database, user, token.GameVersion, token.Platform);
 
         // Do not allow matchmaking if it has been disabled
-        if (!ServerConfiguration.Instance.Matchmaking.MatchmakingEnabled) return this.BadRequest();
+        if (!ServerConfiguration.Instance.Matchmaking.MatchmakingEnabled)
+            if (!ServerConfiguration.Instance.Matchmaking.PatchworkMatchmakingEnabled)
+                return this.BadRequest();
```
Slendy: Can this be cleaned up instead of having 2 nested if statements with no braces.

ProjectLighthouse.Servers.GameServer/Helpers/PatchworkHelper.cs
```diff
@@ -8,17 +8,34 @@ public static partial class PatchworkHelper
     private static readonly int requiredMajor = ServerConfiguration.Instance.Authentication.PatchworkMajorVersionMinimum;
     private static readonly int requiredMinor = ServerConfiguration.Instance.Authentication.PatchworkMinorVersionMinimum;
 
-    [GeneratedRegex(@"^PatchworkLBP[123V] (\d{1,5})\.(\d{1,5})")]
+    // TODO: Ideally move off a regex at some point
```
Slendy, line 11: What's this for? Would the alternative be a custom parser?

ProjectLighthouse/Configuration/ConfigurationCategories/MatchmakingConfiguration.cs
```diff
@@ -3,4 +3,7 @@
 public class MatchmakingConfiguration
 {
     public bool MatchmakingEnabled { get; set; } = true;
+
+    // If `MatchmakingEnabled` is false, this option will still allow matchmaking for Patchwork users
+    public bool PatchworkMatchmakingEnabled { get; set; } = true;
```
Slendy: This needs a config version bump to take effect.

ProjectLighthouse.Servers.GameServer/Controllers/Matching/MatchController.cs
```diff
@@ -45,8 +46,21 @@ public async Task<IActionResult> Match()
...
+        }
+
+        if (ServerConfiguration.Instance.Authentication.RequirePatchworkUserAgent &&
+            token.PatchworkMajor == null || token.PatchworkMinor == null)
+        {
+            return this.BadRequest();
+        }
+
+        // Do not allow users with different join keys to match at all
+        if (token.PatchworkJoinKeyEnabled == true) return this.BadRequest();
```
Slendy, line 63: Shouldn't this functionality also be able to be turned off via the `RequirePatchworkUserAgent` flag?

Slendy (changes requested):
Thanks for your contribution. See comments for more details

ProjectLighthouse.Servers.GameServer/Helpers/PatchworkHelper.cs
FeTetra: Yeah, regex is fine for now, but Refresh for example just manually parses it and it's easier to read the code for overall.

ProjectLighthouse.Servers.GameServer/Controllers/Matching/MatchController.cs
FeTetra: No, if a user connects with Patchwork using a join key (even if the server doesn't enforce its use), they shouldn't be added to the matchmaking pool. This is because the join key system modifies an internal XXTEA network key which handles encryption between P2P clients. Trying to connect to someone who has a different key than you will always fail.

ProjectLighthouse/Types/Entities/Token/GameTokenEntity.cs
FeTetra: 1. I considered not storing the major and minor, but Zaprit told me we might as well do it to keep track of Patchwork version statistics on the Grafana. We're going to have to bump the minimum version at some point, so it will be useful to see how much people are updating before doing so. Also, I do use them in the matchmaking endpoint as the way of checking whether the user has Patchwork.
2. The info is stored in the token simply for convenience. I can let the auth endpoint code handle the initial parsing and then I don't have set up the variables and whatnot to perform these checks in other places in code. It's more of a future-proofing thing.

ProjectLighthouse.Servers.GameServer/Controllers/Login/LoginController.cs
```diff
@@ -213,12 +213,19 @@ public async Task<IActionResult> Login()
             return this.Forbid();
         }
 
-        if (ServerConfiguration.Instance.Authentication.RequirePatchworkUserAgent)
+        bool? hasKey;
+        int? major, minor;
+        bool hasValidPatchworkUserAgent = PatchworkHelper.IsValidPatchworkUserAgent(
+            this.Request.Headers.UserAgent.ToString(), 
+            out major, out minor, out hasKey);
```
Slendy:
```suggestion
        bool hasValidPatchworkUserAgent = PatchworkHelper.IsValidPatchworkUserAgent(
            this.Request.Headers.UserAgent.ToString(), 
            out int? major, out int? minor, out bool? hasKey);
```

FeTetra:
Matchmaking tests appear broken, but I have tested it on my machine and it does pass. I think it could be another weird race condition I've somehow caused. Either way, I've no way to determine what bad codepath it could be going down since I can't debug it on my own machine.

ProjectLighthouse.Servers.GameServer/Controllers/Login/LoginController.cs
Slendy:
```suggestion
        bool hasValidPatchworkUserAgent = PatchworkHelper.IsValidPatchworkUserAgent(
            this.Request.Headers.UserAgent.ToString(), 
            out int? major, out int? minor, out int? hasKey);
```

Slendy (approved):
LGTM, just inline the out variables for the patchwork check in the login function

ProjectLighthouse.Servers.GameServer/Controllers/Login/LoginController.cs
```diff
@@ -213,12 +213,17 @@ public async Task<IActionResult> Login()
             return this.Forbid();
         }
 
-        if (ServerConfiguration.Instance.Authentication.RequirePatchworkUserAgent)
+        bool hasValidPatchworkUserAgent = PatchworkHelper.IsValidPatchworkUserAgent(
+            this.Request.Headers.UserAgent.ToString(), 
+            out int? major, out int? minor, out int? hasKey);
```
Zaprit: Just going off the check failures, does this not make more sense
```suggestion
            out int? major, out int? minor, out bool? hasKey);
```

Zaprit (approved):

PR #1115, opened 2026-08-28
Fix website integration tests inconsistency

Slendy:
Improves random test user generation and adds some extra asserts that make debugging website test failures in CI much more straightforward.


FeTetra:
The patchwork test is failing because the regex will match anything proceeding the user agent, it's a bug I've accidentally introduced, but have since fixed it in #1114.

FeTetra (approved):

PR #1118, opened 2026-08-29
Make /login only create rooms if matchmaking is enabled

FeTetra:
This commit prevents users from creating rooms at authentication to prevent flooding the matchmaking pool with invalid rooms.


Zaprit (approved):

FeTetra:
i was just playing bruh

PR #1120, opened 2026-08-30
Don't block duplicate slot uploads on republishes

FeTetra:
oops


FeTetra:
sorry ads

W0lf4llo:
LGTM! (I cant review)

Slendy (approved):

Slendy (changes requested):
jk actually, shouldn't this use the user ID of the token/user submitting the request and not whatever was included in the request body?

FeTetra:
> jk actually, shouldn't this use the user ID of the token/user submitting the request and not whatever was included in the request body?

That's probably a better idea.

Slendy (approved):

Slendy (approved):

